PR #71, opened 2012-08-07
debian packaging work, attempt 1.

nergdron:
Test and see if this works with LP and making a PPA. :)


luke-jr:
Please merge 50241ff and adapt packaging (ADL_SDK is no longer needed).

luke-jr:
Also, might check to see if this needs to install a file in /etc/modprobe.d to pick up BFLs...

```
ftdi_sio vendor=0x0403 product=0x6014
```

PR #75, opened 2012-08-08
Updated packaging to include modprobe config.

nergdron:
Also documentation changes.


README-Debian
```diff
@@ -5,6 +5,8 @@ Here's the short list on how to build your own Debian/Ubuntu package:
 * Install and configure pbuilder (https://wiki.ubuntu.com/PbuilderHowto)
 * cd into this source directory, and create a fake upstream tarball:
 	tar -cJvf ../bfgminer_${VERSION}.orig.tar.xz *
+* Create the Debian specific bitforce module config from its patch:
```
luke-jr, line 6: Shouldn't pdebuild be doing the patching? O.o

debian/rules
```diff
@@ -4,12 +4,11 @@
...
 override_dh_auto_configure:
-	dh_auto_configure -- --enable-cpumining
-
+	./autogen.sh
+	dh_auto_configure -- --enable-ztex --enable-bitforce --enable-icarus --enable-cpumining
 
 override_dh_auto_install:
-	$(MAKE) DESTDIR=$$(pwd)/debian/bfgminer  install
-
+	$(MAKE) DESTDIR=$(CURDIR)/debian/bfgminer install || sleep 300
```
luke-jr: Why sleep?

nergdron: Whoops, that was a debugging line I left in when it wasn't building, because pdebuild erases the build dir when it fails, and you can't examine it.

README-Debian
nergdron, line 6: It does. The problem comes from when you create a Debian source package from a source directory, it assumes the source directory is post-patching, not pre-patching, and so it tries to reverse the patch before creating the source package. If the file isn't present in the working dir, it fails to reverse the patch, and so fails to create the source patch.

Comes from Debian's idea that you have a pristine source directory that you then add the debian dir and patches to, it's not really well designed around the idea of including the debian dir with the source, unfortunately.

debian/rules
luke-jr: Want to pull it out before I merge?

nergdron: Probably a good idea, but I won't get to it until tomorrow. Up to you.

luke-jr: No problem, no rush.

PR #122, opened 2012-09-03
Formatting fix for GPUs that lack temp/fan data

Sonarman:
My system has an NVidia card, for which get_statline_before returns an empty string, so the status looks like:

```
OCL  0:|  37.9/ 38.2/ 41.2Mh/s | A:14 R:0 HW:0 U: 0.58/m
CPU  0:                |   1.1/  1.2/  0.0Mh/s | A: 0 R:0 HW:0 U: 0.00/m
CPU  1:                |   1.2/  0.7/  0.0kh/s | A: 0 R:0 HW:0 U: 0.00/m
```

This commit prints the proper amount of whitespace in this situation by checking for an empty string.

### DIFF
--- a/miner.c
+++ b/miner.c
@@ -1776,13 +1776,14 @@ static void curses_print_devstatus(int thr_id)
 	if (wmove(statuswin, ypos, 0) == ERR)
 		return;
 	wprintw(statuswin, " %s %*d: ", cgpu->api->name, dev_width, cgpu->device_id);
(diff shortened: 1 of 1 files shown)


luke-jr:
What version are you using? I thought this was fixed in recent versions...

Sonarman:
You're right. It's already fixed. Turns out I was using 2.6.4. My mistake, heh.

Issue #130, opened 2012-09-08
Pool Priority commandline/config, and TUI


purelithium:
Currently, if a user has multiple pools entered in the config file,  the miner uses that order as the pool priority. This requires a full miner restart and time/hashes lost. The user should be able to modify this priority list through the [P]ool Management selection. This modification should then persist through shutdown and restart if the user writes the changes to the config file through the [S]ettings selection.

jaketri:
If pool management code is touched then it would be nice to be able to save/restore from/to config file the pool state (enable/disable).

luke-jr:
@jaketri That sounds like a slightly more in-depth change, but really should be in another issue of its own.

jaketri:
@luke-jr OK no problem ... I'll open a separate issue for it ... I was thinking it is manage by same code ...

luke-jr:
@blinkier implemented this

PR #153, opened 2012-09-28
Issue #130 / Pool Reorder

blinkier:
Pools can be reordered/reprioritized through UI

### DIFF
--- a/miner.c
+++ b/miner.c
@@ -3411,19 +3411,11 @@ int prioritize_pools(char *param, int *pid)
 	int i, pr, prio = 0;
 
 	if (total_pools == 0) {
-<<<<<<< HEAD
 		return MSG_NOPOOL;
 	}
 
 	if (param == NULL || *param == '\0') {
 		return MSG_MISPID;
-=======
-		return 8; //MSG_NOPOOL
-	}
-
-	if (param == NULL || *param == '\0') {
-		return 25; //MSG_MISPID
->>>>>>> 2ad08ed2b80fbb4810a569e5ea26ca311b6c894c
 	}
 
 	bool pools_changed[total_pools];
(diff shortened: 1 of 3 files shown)


miner.c
```diff
@@ -4247,6 +4247,39 @@ static void display_pools(void)
...
+                        goto retry;
+                }
+                selected = curses_int("Select pool number");
+                if (selected < 0 || selected >= total_pools) {
+                        wlogprint("Invalid selection\n");
+                        goto retry;
+                }
+                priority = curses_int("Select new priority");
+                if (priority < 0 || priority >= total_pools) {
+                        wlogprint("Invalid selection\n");
```
luke-jr: Perhaps "Invalid priority\n"?

luke-jr:
This seems to actually try to move the pool in the pools array. That will cause all sorts of problems. Only pool->prio should be adjusted. Perhaps check out the start of switch_pools?

blinkier:
Yes, it does reorder the pool array. The idea is to be able to then write the new order into a config file (since this is interpreted as priority on read in). The only thing that's perhaps counter intuitive is that if you give a pool priority 0, bfgminer doesn't immediately switch to it.

I've been using the code since 2.7.5 and haven't really had any issues with it. I looked, very briefly, at the pool switching strategies and didn't immediately see any problems. Did I miss something?

miner.c
```diff
@@ -4247,6 +4247,39 @@ static void display_pools(void)
...
+                        for (i = selected; i < priority; i++)
+                                pools[i] = pools[i + 1];
+                } else {
+                        for (i = selected; i > priority; i--)
+                                pools[i] = pools[i - 1];
+                }
+                pools[priority] = pool;
+		for (i = 0; i < total_pools; i++)
+			pools[i]->pool_no = pools[i]->prio = i; 
+
```
blinkier: Perhaps this makes sense?

```
if (selected == 0 || priority == 0)
    switch_pools(pools[0]);
```

luke-jr:
Probably it would make sense to abstract and base this on the RPC poolpriority command (api.c:2189)

Issue #167, opened 2012-10-16
BFGMiner crash with access violation


jaketri:
Access violation when starting bfgminer on AMD 79xx series with thread-concurrency set to 24576.

PR #168, opened 2012-10-16
Fix for issue #167 / BFGMiner crash with access violation

jaketri:
This is a fix for issue #167

First, the access violation was cause by line #721 from ocl.c where "numbuf" was to small for any thread-concurrency > 9999. To fix it I just increased the size of numbuf from 10 chars to 32.
Another possible way to fix it is to split the operation from lines #721-722 into 2 , first process lookup_gap then thread_concurrency.

After that I was able to use thread_concurrency up to 16384. For more memory allocation for padbuffer8 buffer failed. Failure is due to logic  that supposed to find the smallest power of 2 value >= required amount of memory. Old logic works fine only if board have memory size a power of 2. The new logic just find the smallest power of 2 value and if it is small enough (less than board memory size) it use it.

### DIFF
--- a/ocl.c
+++ b/ocl.c
@@ -557,7 +557,7 @@ _clState *initCl(unsigned int gpu, char *name, size_t nameSize)
 	 */
 	char binaryfilename[255];
 	char filename[255];
-	char numbuf[10];
+	char numbuf[32];
 
 	if (cgpu->kernel == KL_NONE) {
 		if (opt_scrypt) {
--- a/ocl.c
+++ b/ocl.c
@@ -652,7 +652,6 @@ _clState *initCl(unsigned int gpu, char *name, size_t nameSize)
 #ifdef USE_SCRYPT
 	if (opt_scrypt) {
 		cl_ulong ma = cgpu->max_alloc, mt;
-		int pow2 = 0;
 
 		if (!cgpu->opt_lg) {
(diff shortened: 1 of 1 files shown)


ocl.c
```diff
@@ -676,12 +675,10 @@ _clState *initCl(unsigned int gpu, char *name, size_t nameSize)
 		 * >= required amount to map nicely to an intensity */
 		mt = cgpu->thread_concurrency * 32768 * cgpu->lookup_gap;
 		if (ma > mt) {
-			while (ma >>= 1)
-				pow2++;
```
luke-jr, line 680: Any reason to prefer this over pow(2, floor(log2(ma, 2))) ?

jaketri, line 680: pow(2, floor(log2(ma, 2)))  will give you the biggest power of 2 value that fit in to full memory instead of minimum one. If you do this it will still work but you will be always limited to one GPU thread all the time because there is no more memory available for another thread.

A possible alternate solution is to use 

```
ma = pow(2, ceil(log2(mt, 2)))
```

instead of 

```
ma = 1;
while (ma < mt)
    ma <<=1;
```

and you will get same result but I guess it is harder to understand by an average developer.

luke-jr:
Looks okay and shouldn't hurt Bitcoin (I don't care much for scrypt) :)

PR #171, opened 2012-11-01
Some x6500 improvements

wizkid057:
After testing out the existing MMQ bitstream and the normal x6500 overclocker bitstream, the x6500 bitstream yields much more stable results.  Far fewer HW errors across the board and hashrates within a few percent of expected right away.

I don't know the exact differences between the bitstreams, but, there is an obvious performance difference.

Also, implement the fix Luke-jr proposed for the HW error loop issue.


wizkid057:
Added dual temp sensor support. in a8069e1

luke-jr:
Please pull out "Implement suggested fix for hardware error loop issue" since I already have that here...

Also need some doc file on the bitstream, at least including its license; see existing bitstreams/ files for example. Or maybe it just needs renaming to fit the fpgaminer file?

wizkid057:
You concerns have been addressed.  For public note, the x6500 bitstream is the same bitstream as the MMQ bitstream compiled with alternate optimizations.

PR #193, opened 2012-12-08
Properly compute the best share (endianness issues)

mbevand:
See https://bitcointalk.org/index.php?topic=61731.msg1386248#msg1386248


miner.c
```diff
@@ -2392,7 +2393,9 @@ bool regeneratehash(const struct work *work)
 	swap32yes(swap32, data32, 80 / 4);
 
 	sha2(swap, 80, hash1, false);
-	sha2(hash1, 32, (unsigned char *)(work->hash), false);
+	sha2(hash1, 32, hash2, false);
+
+	swap32yes((unsigned char *)work->hash, hash2, 32 / 4);
```
luke-jr, line 2398: IMO, work->hash should be expected to hold the byte-for-byte SHA256 result.

luke-jr:
Could you review these two fixes? Thanks!
For 2.10.0, rewriting the problem functions in a more readable/understandable way: bfab076d593978c851e712de5a7fadd39173465c (ignore accidental libblkmaker change)
For 2.8.x and 2.9.x, simply fixing the main problem of work->hash being flipped around: 006faac674358b0cc5e346f943d658db488d16d2

luke-jr:
Please close when this is confirmed fixed, thanks.

PR #201, opened 2013-01-04
Update windows-build.txt

bluemurder:
Added jansson library installation steps

### DIFF
--- a/windows-build.txt
+++ b/windows-build.txt
@@ -139,7 +139,17 @@ Copy src/.libs/*.dll into \MinGW\bin
 Copy src/.libs/*.a, src/.libs/*.la, src/.libs/*.lai into \MinGW\lib
 Copy src/jansson.def into \MinGW\lib
 Copy src/*.h into \MinGW\include
-Copy jansson.pc into \MinGW\lib\pkgconfig
+Create jansson.pc into \MinGW\lib\pkgconfig with this content:
+
+prefix=.
+exec_prefix=${prefix}/bin
+libdir=${exec_prefix}/lib
+includedir=${prefix}/include
+Name: jansson
+Description: Library for encoding, decoding and manipulating JSON data
+Version: 2.4
+Libs: -L${libdir} -ljansson
+Cflags: -I${includedir} @CPPFLAG_JANSSON_STATICLIB@
 
 (Source: http://www.digip.org/jansson/doc/dev/gettingstarted.html)
 (Source: https://groups.google.com/forum/#!msg/jansson-users/CTPFBeZ0ZWQ/W3uuOuWZegkJ)
--- a/windows-build.txt
+++ b/windows-build.txt
@@ -138,7 +138,7 @@ Enter jansson-X.X dir, and type
 Copy src/.libs/*.dll into \MinGW\bin
 Copy src/.libs/*.a, src/.libs/*.la, src/.libs/*.lai into \MinGW\lib
 Copy src/jansson.def into \MinGW\lib
-Copy src/*.h into \MinGW\include\jansson
+Copy src/*.h into \MinGW\include
 Copy jansson.pc into \MinGW\lib\pkgconfig
 
 (Source: http://www.digip.org/jansson/doc/dev/gettingstarted.html)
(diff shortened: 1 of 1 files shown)


windows-build.txt
```diff
@@ -123,6 +123,38 @@ Edit the file \MinGW\lib\pkgconfig\libcurl.pc and change "-lcurl" to
...
+(Start Icon/keyboard key ==> All Programs ==> MinGW ==> MinGW Shell).
+Enter jansson-X.X dir, and type
+      ./configure
+      make
+      make check
+      make install
+Copy src/.libs/*.dll into \MinGW\bin
+Copy src/.libs/*.a, src/.libs/*.la, src/.libs/*.lai into \MinGW\lib
+Copy src/jansson.def into \MinGW\lib
+Copy src/*.h into \MinGW\include
```
luke-jr, line 141: Shouldn't all these be done by "make install"? Does that not work correctly on MingW?

windows-build.txt
```diff
@@ -123,6 +123,38 @@ Edit the file \MinGW\lib\pkgconfig\libcurl.pc and change "-lcurl" to
...
+
+prefix=.
+exec_prefix=${prefix}/bin
+libdir=${exec_prefix}/lib
+includedir=${prefix}/include
+Name: jansson
+Description: Library for encoding, decoding and manipulating JSON data
+Version: 2.4
+Libs: -L${libdir} -ljansson
+Cflags: -I${includedir} @CPPFLAG_JANSSON_STATICLIB@
```
luke-jr, line 152: Does this actually work? .pc files shouldn't have these @VAR@ type things in them, usually.

windows-build.txt
bluemurder, line 141: make install run in a successful way but when configuring bfgminer a "libjansson not found" appears..

luke-jr, line 141: Is it perhaps putting the files in a different place than bfgminer expects? What if you add --prefix=/MinGW/ to the ./configure options?

windows-build.txt
bluemurder, line 152: I tried (with success) this configuration yesterday when building for the first time bfgminer with mingw. I wrote the @CPPFLAG_JANSSON_STATICLIB@ keyword observing the libcurl.pc obtained from http://curl.haxx.se/gknw.net/7.28.1/dist-w32/curl-7.28.1-devel-mingw32.zip that is as follows:

prefix=.
exec_prefix=${prefix}/bin
libdir=${prefix}/lib
includedir=${prefix}/include
supported_protocols="dict file ftp ftps gopher http https imap imaps ldap pop3 pop3s rtmp rtsp scp sftp smtp smtps telnet tftp"
supported_features="AsynchDNS GSS-Negotiate IDN Largefile NTLM SPNEGO SSL SSPI libz"

Name: libcurl
URL: http://curl.haxx.se/
Description: Library to transfer files with ftp, http, etc.
Version: 7.28.1
Libs: -L${libdir} -lcurl -lcurldll
Libs.private:
Cflags: -I${includedir} @CPPFLAG_CURL_STATICLIB@

windows-build.txt
bluemurder, line 141: --prefix=/MinGW/ can be the simpler colution. I published first working solution for my case, following instructions in https://groups.google.com/forum/#!msg/jansson-users/CTPFBeZ0ZWQ/W3uuOuWZegkJ

windows-build.txt
luke-jr, line 152: Ironically, that libcurl binary is broken and bfgminer wouldn't build with it except that I have a workaround to remove the @CPPFLAG_CURL_STATICLIB@ bit.

libjansson should make and install its own .pc file if you get the 'make install' working correctly, I think

luke-jr:
Ok, if this worked for you, let's just merge it. If you can figure out a simpler way (I think the configure prefix might make everything else just work), it'd be nice to add another commit to update it.

GilesBathgate:
@luke-jr you are right, libcurl binary releases have @CPPFLAG_CURL_STATICLIB@ in the package config .pc file which is a [known bug](http://sourceforge.net/p/curl/bugs/1130/). When you build from source the libcurl.pc will contain the define -DCURL_STATICLIB. Putting @CPPFLAG_JANSSON_STATICLIB@ is completely wrong even if something translated this into -DJANSSON_STATICLIB then it would still be wrong because there is no such define for libjansson. I had to remove it to get this to compile on winxp, and [several others](https://bitcointalk.org/index.php?topic=170260.0) are having the same problem.

PR #211, opened 2013-04-03
Translated README to MarkDown format

suranyami:
It reads better on the Github repo landing page. 
It's a little thing, and entirely optional, but I found it useful.

Feel free to ignore it, but use it if you find it useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,1089 @@
+This is a multi-threaded multi-pool ASIC, FPGA, GPU and CPU miner with ATI GPU
+monitoring, (over)clocking and fanspeed support for bitcoin and derivative
+coins. Do not use on multiple block chains at the same time!
+
+This code is provided entirely free of charge by the programmer in his spare
+time so donations would be greatly appreciated. Please consider donating to the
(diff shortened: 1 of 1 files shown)


luke-jr:
Is it possible to generate the README plaintext from this?

suranyami:
Initial investigation says 'yes'. Just gonna check if there's a way to do it from a clean linux install. (i.e. doesn't depend on installing some other lib just for this).

luke-jr:
I've not found a simple way to do MD->TXT - other dependencies don't matter (I'll just make it part of the pre-build process).

suranyami:
# I did a bit of hunting on the weekend and couldn't find an easy way either.

yskywalker:
Why not include a markdown and a plain text version.  I think it's worth the extra work of keeping both updated because being able to add formatting makes a big difference in the clarity, simplicity and attractiveness of the repo main page.

luke-jr:
This is now obsolete; please feel free to reopen with an updated version if we can autogenerate txt from md.

PR #213, opened 2013-04-05
fixed typo in api-example.c, added SPEC file for SUSE distributions, added Python script using RPC API

berendt:


luke-jr:
For new versions, should I just bump the version in the spec file?

PR #217, opened 2013-04-09
updated SUSE SPEC to version 3.0.0

berendt:


packaging/suse/Makefile.am.patch
```diff
@@ -1,11 +1,11 @@
...
+@@ -38,7 +38,7 @@
  if NEED_LIBBLKMAKER
  SUBDIRS           += libblkmaker
  bfgminer_CPPFLAGS += -Ilibblkmaker
 -bfgminer_LDFLAGS  += -Llibblkmaker/.libs -Wl,-rpath,libblkmaker/.libs
 +bfgminer_LDFLAGS  += -Llibblkmaker/.libs
- bfgminer_LDADD    += -lblkmaker-0.1 -lblkmaker_jansson-0.1
- 
+ bfgminer_LDADD    += -lblkmaker_jansson-0.1 -lblkmaker-0.1
+
```
berendt, line 10: The order of lblkmaker-0.1 and lblkmaker_jansson-0.1 changed in the delivered Makefile.am. Because I updated the patch file the order in the patch has also to be adjusted. This will not change the order of the libraries in Makefile.am.

PR #243, opened 2013-06-21
Update driver-avalon.c

aktalak:
allows to overclock the avalon up to 375 mhz per chip!


luke-jr:
Merged via d1e5435acf5d390b97dd5029797b356a816f5570

PR #246, opened 2013-06-27
Fixes bug with ncurses and switching debug-logging into syslog-facility

MaStr:


luke-jr:
Pretty sure this breaks the usual use-case of --debuglog...?

luke-jr:
Feel free to resubmit a new pullreq if you decide to go forward with this...

PR #253, opened 2013-07-10
fix infinite loop in bytes_cpy when size is zero

iongchun:
Beside this bug, I don't know if update_last_work() should call copy_work with zero-sized nonce2.

### DIFF
--- a/util.h
+++ b/util.h
@@ -182,7 +182,7 @@ void bytes_cpy(bytes_t *dst, const bytes_t *src)
 	dst->allocsz = src->allocsz;
 	dst->sz = src->sz;
 	size_t half;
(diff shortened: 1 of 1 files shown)


luke-jr:
It shouldn't, but you're probably right this should be fixed in case bytes_t ends up used elsewhere.

Can you do the check within the loop so you can just return early? malloc(0) doesn't work on all platforms, and there's no data to copy in this case...

iongchun:
I prefer to return early before the loop ;p
Just trace the bytes_\* functions a little more, because the codes managing bytes_t.buf outside these functions makes me uneasy.
Maybe I am too far away from C-style...

luke-jr:
Looks good.

PR #256, opened 2013-07-23
Using primecoin addresses for --coinbase-addr

SoftwareMensch:
Is it possible the change the code in libblmaker/base58.c line 120 to:

```
    switch (addrver) {
            case   0:  // Bitcoin pubkey hash
            case  23:  // Primecoin pubkey hash
            case 111:  // Testnet pubkey hash
                    if (outsz < (rv = 25))
                            return rv;
                    cout[ 0] = 0x76;  // OP_DUP
                    cout[ 1] = 0xa9;  // OP_HASH160
                    cout[ 2] = 0x14;  // push 20 bytes
                    memcpy(&cout[3], &addrbin[1], 20);
                    cout[23] = 0x88;  // OP_EQUALVERIFY
                    cout[24] = 0xac;  // OP_CHECKSIG
                    return rv;
            case   5:  // Bitcoin script hash
            case  83:  // Primecoin script hash
            case 196:  // Testnet script hash
                    if (outsz < (rv = 23))
                            return rv;
                    cout[ 0] = 0xa9;  // OP_HASH160
            default:
                    return 0;
    }
```

See: http://pastebin.com/XSTTRNF9

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -115,6 +115,9 @@ bfgminer_SOURCES += scrypt.c scrypt.h
 dist_doc_DATA += README.scrypt
 endif
 
+bfgminer_SOURCES += prime.c
+bfgminer_LDADD += -lgmp
+
 if HAS_CPUMINE
 bfgminer_SOURCES	+= \
 		  sha256_generic.c sha256_via.c	\
--- a/Makefile.am
+++ b/Makefile.am
@@ -117,6 +117,7 @@ endif
 
 bfgminer_SOURCES += prime.c
 bfgminer_LDADD += -lgmp
+bfgminer_CPPFLAGS += -DWANT_PRIMEPOW
 
 if HAS_CPUMINE
 bfgminer_SOURCES	+= \
--- a/Makefile.am
+++ b/Makefile.am
@@ -117,7 +117,7 @@ endif
 
 bfgminer_SOURCES += prime.c
 bfgminer_LDADD += -lgmp
-bfgminer_CPPFLAGS += -DWANT_PRIMEPOW
+bfgminer_CPPFLAGS += -DWANT_PRIMEPOW -DUSE_WEAVE_CHEMISIST
 
 if HAS_CPUMINE
 bfgminer_SOURCES	+= \
--- a/driver-cpu.c
+++ b/driver-cpu.c
@@ -132,6 +132,8 @@ extern bool scanhash_sse2_32(struct thr_info*, const unsigned char *pmidstate, u
 	uint32_t max_nonce, uint32_t *last_nonce,
 	uint32_t nonce);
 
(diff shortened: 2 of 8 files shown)


luke-jr:
No.

SoftwareMensch:
I know that this library is for bitcoin, but bfgminer is great, also for primecoinming. We need a official solution.

luke-jr:
The solution is to not use scamcoins.

SoftwareMensch:
This is an interesting point of view. Why do you call it scamcoins? I like the idea using primes. Please give me facts which confirms your statement.

luke-jr:
The concept of a prime proof-of-work, while interesting and noteworthy, is just an algorithm. A cryptocurrency is built around economics. Mining is not the purpose of cryptocurrencies, just an implementation detail - and the POW used for it is yet another deeper layer of technicality.

With a view of the big picture, Primecoin is just a Bitcoin clone with nothing _notable_ different. As a result, we see Primecoin being used as a mere vehicle to pump-and-dump and "make a profit", but without any actual reason to use it as a currency.

While I still see value in the prime POW from a research perspective, I don't see any good from support for _Primecoin_ itself.

SoftwareMensch:
You're right, all the cryptocoins, including litecoin and so on, are clones
from bitcoin. Bitcoin is the big brother and it needs a long time to find a
better one, so far it is possible. For primecoin, the main advantage imho
is the using of prime numbers and the extremely fast transactions. And yes
the possibility to mine some stuff. The difficulty of bitcoin makes no
sense to mine, because all the asic developer don't want to sell their
hardware, they use it by their own.
Am 24.07.2013 20:40 schrieb "Luke-Jr" notifications@github.com:

luke-jr:
Increasing block rate may in a vacuum make transactions faster, but in the end, Primecoin transactions are in fact slower than Bitcoin.

All ASIC designers are selling hardware. Only one of them (Bitfountain) is doing any notable self-mining.

SoftwareMensch:
> Increasing block rate may in a vacuum make transactions faster, but in
> the end, Primecoin transactions are in fact slower than Bitcoin.

Good to know.

> All ASIC designers are selling hardware. Only one of them (Bitfountain)
> is doing any notable self-mining.

I think they selling little pieces while more is possible. In past I tried
to buy an avalon asic. An absolutly nightmare.

luke-jr:
They're certainly backordered quite a bit. But while more might be possible, I don't think it's intentionally slow.

SoftwareMensch:
Sure? If you're thinking business like, what would you do? Sell the stuff
on mass which is earning money for you, or use it by your own and tell the
world that you only have a limited batch? This behavior is natural, its
human.
Am 24.07.2013 21:25 schrieb "Luke-Jr" notifications@github.com:

luke-jr:
The people who made it rich the most during the gold rush were the companies that just sold shovels.

SoftwareMensch:
Nice comparison. But with a couple oft some 20 ghash/s monster from bfl,
the profit margin is extremely great.

But in fact, I understand why you dislike such clone. But why is there a
bfgminer branch called "prime"?
Am 24.07.2013 21:39 schrieb "Luke-Jr" notifications@github.com:

PR #265, opened 2013-08-13
add rpc cpu enable/disable/restart

lapoigne:
Hello,

I added commands for cpuenable cpudisable and cpurestart on the rpc interface.


luke-jr:
Any chance you'd be interested in going further to abstract away CPU/GPU to use the same pga\* commands as everything else? :)

lapoigne:
I'll look at that when I'll have at bit of time.

PR #267, opened 2013-08-21
Changed comparison constants to allow for floating-point rounding

Krellan:
Here is a proposed fix for the floating-point rounding issue I saw that produced the (.100%) display which overflowed the column width by 1 character.  Tested locally on my mining rig and seems to work.


luke-jr:
Thanks!

Krellan:
Cool, you're welcome!

PR #279, opened 2013-09-13
Support for bitfurystrikesback boards

anajavi:
This is a quick hack to make bfgminer to work on bitfurystrikesback boards.

Works mostly the same as metabank boards, but spi speed is lower due to longer chains of chips.

This driver actually should not be autodetected as it is incompatible with the megabigpower boards.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -211,7 +211,7 @@ bfgminer_SOURCES += driver-ztex.c libztex.c libztex.h
 endif
 
 if HAS_BITFURY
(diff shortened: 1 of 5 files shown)


luke-jr:
noserial_detect_manual actually only calls autodetect iff the user explicitly requests it with -S <dname>:auto

Note that the bitfury code (including both branches) is still under active development, and I don't have a sample board yet, so I can't guarantee something won't break in git as it progresses. If you could yell at me when/if I break it (perhaps join the devel ML so you get advance warning before a release), I'd appreciate it. :)

anajavi:
Thanks.

I will ping you if it breaks:)
I Need to still work with timing issues, I guess these boards need different timing than the Metabank ones.

PR #285, opened 2013-09-21
Change in struct bitfury_device

anajavi:
This change makes driver to send less duplicates.
Also it makes it a lot easier to do my own scanwork in driver-bfsb.c.

I didn't change littlefury driver because I can't test it. I can write that one without testing if you're okay with it.

Thanks.


luke-jr:
Yes, please update littlefury's driver too.

PR #286, opened 2013-09-23
Allow to build only *fury drivers

alexxy:
Also reoreder output options

Signed-off-by: Alexey Shvetsov alexxy@gentoo.org


luke-jr:
Please use a separate commit for each change. Thanks!

PR #290, opened 2013-09-27
Bitfury

alexxy:


luke-jr:
Why not just /etc/bfgminer.conf ? Also, this should use autoconf-provided path instead of hard-coding /etc
Finally, I'm not sure it's good to encourage running as root. Maybe /etc/bfgminer.conf should be loaded regardless of user (in addition to user's own)?

PR #291, opened 2013-09-27
Added get_api_extra_device_status for 'devs' request in metabank driver: Slot, Clock Bits, Temperature, Voltage.

Viktor-KN:


luke-jr:
Please store temperatures on cgpu_info.temp instead.

Viktor-KN:
Done.

luke-jr:
This is reporting 35.2 C for both slots, but I'd expect much hotter from physical touch...

I think i2c is not threadsafe, so metabank_get_stats will need to use a mutex possibly. But maybe I can make it call from the same thread - will get back to you on that :)

Viktor-KN:
It reporting from sensor on the board, not from chips itself. Sensor is located in not very good place... But this is all we have...

luke-jr:
Thanks

PR #293, opened 2013-09-29
chroot support

corrideat:
I've managed to get some pretty basic chroot functionality working.

I believe this is very useful to run the miner as an isolated a process as possible, and would love to see this feature in a future release.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -74,6 +74,14 @@ AC_CHECK_HEADERS([sys/epoll.h])
 AC_CHECK_HEADERS([sys/prctl.h])
 AC_CHECK_HEADERS([sys/file.h])
 
+AC_CHECK_HEADERS([sys/file.h])
+
+# Setuid
+AC_CHECK_HEADERS([pwd.h])
+
+# Check for chroot support
+AC_CHECK_FUNCS([chroot])
+
 AC_FUNC_ALLOCA
 
 have_cygwin=false
@@ -314,20 +322,20 @@ AC_ARG_ENABLE([icarus],
 	[icarus=$enableval],
 	[icarus=yes]
 	)
-AM_CONDITIONAL([HAS_ICARUS], [test x$icarus = xyes])
 if test "x$icarus" = xyes; then
 	AC_DEFINE([USE_ICARUS], [1], [Defined to 1 if Icarus support is wanted])
 fi
+AM_CONDITIONAL([HAS_ICARUS], [test x$icarus = xyes])
 
-chroot = "no"
+chroot = "yes"
 
 AC_ARG_ENABLE([chroot],
-	[AC_HELP_STRING([--disable-chroot],[Compile support for running inside a chroot jail])],
+	[AC_HELP_STRING([--disable-chroot],[Compile support for running inside a chroot jail (default enabled)])],
 	[chroot=$enableval],
 	[chroot=yes]
 	)
-if test "x$chroot" = xchroot; then
-	AC_DEFINE([CHROOT], [1], [Defined to 1 if chroot jail support is wanted])
+if test "x$chroot" = xyes; then
+	AC_DEFINE([BFG_CHROOT], [1], [Defined to 1 if chroot jail support is wanted])
 fi
 
 avalon="no"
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -314,10 +314,21 @@ AC_ARG_ENABLE([icarus],
 	[icarus=$enableval],
 	[icarus=yes]
 	)
+AM_CONDITIONAL([HAS_ICARUS], [test x$icarus = xyes])
 if test "x$icarus" = xyes; then
 	AC_DEFINE([USE_ICARUS], [1], [Defined to 1 if Icarus support is wanted])
 fi
-AM_CONDITIONAL([HAS_ICARUS], [test x$icarus = xyes])
```
luke-jr, line 320: Why is this changing Icarus code?

configure.ac
```diff
@@ -314,10 +314,21 @@ AC_ARG_ENABLE([icarus],
...
+
+chroot = "no"
+
+AC_ARG_ENABLE([chroot],
+	[AC_HELP_STRING([--disable-chroot],[Compile support for running inside a chroot jail])],
+	[chroot=$enableval],
+	[chroot=yes]
+	)
+if test "x$chroot" = xchroot; then
+	AC_DEFINE([CHROOT], [1], [Defined to 1 if chroot jail support is wanted])
```
luke-jr: Please rename this to BFG_CHROOT or similar, in case some library we use looks for CHROOT in future/other-os headers.

configure.ac
corrideat, line 320: I've just realised this. I supposed I messed up when editing the configure.ac.

configure.ac
```diff
@@ -314,10 +314,21 @@ AC_ARG_ENABLE([icarus],
...
 	AC_DEFINE([USE_ICARUS], [1], [Defined to 1 if Icarus support is wanted])
 fi
-AM_CONDITIONAL([HAS_ICARUS], [test x$icarus = xyes])
+
+chroot = "no"
+
+AC_ARG_ENABLE([chroot],
+	[AC_HELP_STRING([--disable-chroot],[Compile support for running inside a chroot jail])],
+	[chroot=$enableval],
+	[chroot=yes]
```
luke-jr, line 335: What about systems that don't support chroot? This should check if it is supported by default, so as to avoid compile errors.

miner.c
```diff
@@ -1474,6 +1482,14 @@ static char *set_null(const char __maybe_unused *arg)
 		     set_int_0_to_9999, opt_show_intval, &opt_bench_algo,
 		     opt_hidden),
 #endif
+#ifdef CHROOT
+        OPT_WITH_ARG("--chroot-dir",
+                     opt_set_charp, NULL, &chroot_dir,
+                     "Chroot to a directory right after startup"),
+        OPT_WITH_ARG("--chroot-user",
+                     opt_set_charp, NULL, &chroot_user,
+                     "Username of an unprivileged user to run as"),
```
luke-jr: Shouldn't this be --setuid?

miner.c
```diff
@@ -10047,6 +10063,29 @@ int main(int argc, char *argv[])
...
 
+#ifdef CHROOT
+        if (chroot_dir != NULL) {
+                struct passwd *user_info = NULL;
+                if (chroot_user != NULL) {
+                        if ((user_info = getpwnam(chroot_user)) == NULL) {
+                                quit(1, "Unable to find user information");
+                        }
+                } else if (getuid() == 0) {
+                        quit(1, "Running as root is not allowed");
```
luke-jr: Currently, running as root must be supported for some GPIO interfaces.

miner.c
corrideat: Possibly. In that case, I would expect a slightly different behaviour though, for it should work independently of --chroot-dir.

miner.c
corrideat: Note that I'm only disabling running as root when chroot is also used, but I'll fix it to just log a warning.

corrideat:
Updated as requested

luke-jr:
Any objection if I remove the compile option, and always build with --chroot-dir support if available?

corrideat:
None at all :)

luke-jr:
Merged, thanks.

PR #295, opened 2013-10-03
fix false switching job_switched

alpet83:
Sometimes bits on MISO in Metabank devices lost, and buffer consist abnormal values.

### DIFF
--- a/libbitfury.c
+++ b/libbitfury.c
@@ -448,7 +448,10 @@ void libbitfury_sendHashData1(int chip_id, struct bitfury_device *d, struct thr_
 			}
 		}
 
-		d->job_switched = newbuf[16] != oldbuf[16];
+		  if ( !newbuf[16] || 0xffffffff == newbuf[16] ) // data from SPI can be wrong
+             		d->job_switched = ( newbuf[16] != oldbuf[16]  ); 
+        	else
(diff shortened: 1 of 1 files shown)


luke-jr:
d->job_switched should be 0 in this case, not left unchanged, no?

Also, this code is no longer used and will probably be removed soon... Is there still a need for it?

alpet83:
job_switched really changes about one time per second. So code can test values more times per second, and react on true switch. Also you can count the same bits if they are less than 16, the flag should be set.  What alternative detection do you use to change work?

luke-jr:
Using a similar check in bitfury_do_io now, but after checking for SPI corruption (comparing the previous nonce).

alpet83:
Have like solution in https://github.com/alpet83/bfgminer/blob/littlefury/libbitfury.c:811 but check only 16 nonces, and job_switch check separately. Trying to run you last bitfury branch on metabank device, but no chips detected... need new arguments in command line?

luke-jr:
Latest code needs --enable-metabank for configure, and -S metabank:auto for runtime.

alpet83:
Ok. Now it's worked! Please correct help for configure: "--disable-metabank      Compile support for Metabank (default 
disabled)"

PR #299, opened 2013-10-09
Link libblkmaker statically

dzindra:
Is there any specific reason for libblkmaker being linked dynamically and in specific path? If there is no such reason, I think it would be better to link it statically. 

In this case there is also a minor change in libblkmaker's configure.ac file:

``` patch
diff --git a/configure.ac b/configure.ac
index 8102143..6e7fc70 100644
--- a/configure.ac
+++ b/configure.ac
@@ -15,7 +15,7 @@ AM_INIT_AUTOMAKE([1.10 -Wall dist-bzip2])

 AC_PROG_CC_C99
 m4_ifdef([AM_PROG_AR], [AM_PROG_AR])
-LT_INIT([disable-static])
+LT_INIT()

 # http://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
 AC_SUBST([LIBBLKMAKER_SO_VERSION], [4:1:4])
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -620,21 +620,9 @@ if test "x$with_system_libblkmaker" = "xyes"; then
 		AC_MSG_ERROR([Could not find system libblkmaker])
 	])
 else
-	save_LDFLAGS="$LDFLAGS"
-	LDFLAGS="$LDFLAGS -Wl,-zorigin"
-	origin_LDFLAGS=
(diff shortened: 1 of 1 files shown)


luke-jr:
Because static linking is always wrong. See http://www.akkadia.org/drepper/no_static_linking.html for an overview..

The specific path is so you can run bfgminer from the build directory without installing it. It's just searched in addition to your OS's usual library paths, so it works fine if you install it too.

PR #303, opened 2013-10-19
Added frequency calculation for bitfury miner

legkodymov:
- Changed copyright for tm_i2c.c.
- Added frequency calculation for bitfury miner.

### DIFF
--- a/driver-bitfury.c
+++ b/driver-bitfury.c
@@ -706,7 +706,7 @@ struct api_data *bitfury_api_device_status(struct cgpu_info * const cgpu)
 	int clock_bits = bitfury->osc6_bits;
 	
 	root = api_add_int(root, "Clock Bits", &clock_bits, true);
-	root = api_add_double(root, "Mhz", &bitfury->mhz, false);
+	root = api_add_freq(root, "Frequency", &bitfury->mhz, false);
 	
 	return root;
 }
--- a/driver-bitfury.c
+++ b/driver-bitfury.c
@@ -680,7 +680,7 @@ out:
 		}
 	}
 	
-	timer_set_delay_from_now(&master_thr->tv_poll, 200000);
+	timer_set_delay_from_now(&master_thr->tv_poll, 10000);
 }
 
 int64_t bitfury_job_process_results(struct thr_info *thr, struct work *work, bool stopping)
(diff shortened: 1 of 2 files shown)


driver-bitfury.c
```diff
@@ -653,7 +680,7 @@ void bitfury_do_io(struct thr_info * const master_thr)
 		}
 	}
 	
-	timer_set_delay_from_now(&master_thr->tv_poll, 10000);
+	timer_set_delay_from_now(&master_thr->tv_poll, 200000);
```
luke-jr: Is slowing this down necessary?

legkodymov: Not necessary. But doing more than 2 polls per "job period" is not necessary. Even 1.1-1.9 polls is enough. I know, p2pool is special case.
I guess it needs special automatic calculation in future or config option.

luke-jr: Even normal pools benefit from more frequent polling (fewer stale shares), just much less. If there's no benefit to polling less often, it probably makes sense to just leave it every 10ms?

driver-bitfury.c
```diff
@@ -679,6 +706,7 @@ struct api_data *bitfury_api_device_status(struct cgpu_info * const cgpu)
 	int clock_bits = bitfury->osc6_bits;
 	
 	root = api_add_int(root, "Clock Bits", &clock_bits, true);
+	root = api_add_double(root, "Mhz", &bitfury->mhz, false);
```
luke-jr: Please change this to:

``` C
root = api_add_freq(root, "Frequency", &bitfury->mhz, false);
```

driver-bitfury.c
```diff
@@ -599,7 +602,31 @@ void bitfury_do_io(struct thr_info * const master_thr)
...
-		
+
+		counter = bitfury_decnonce(newbuf[n]);
+		if ((counter & 0xFFC00000) == 0xdf800000)
+		{
+			unsigned int cycles;
+			long long unsigned int period;
+			double ns;
+			struct timeval d_time;
+			static int skip;
```
luke-jr, line 613: Sorry, I missed this: why is it static?

luke-jr:
Merged as 75a7e14

PR #304, opened 2013-10-23
revert change to .gitignore and add framework to modify bitfury osc6 on startup (range: 45-55)

TrooperT:


luke-jr:
My plan was to add a --set-device NFY:osc6_bits=50 or similar generic option. Thoughts?

TrooperT:
The formatting seems a bit odd to me but tis my two satoshi.

Like I said it's just an initial idea.

Also the .gitignore change is to undo an accident in the previous commit I did

PR #308, opened 2013-10-25
dynamic osc_6 bits, updated

asfin:


driver-bitfury.c
```diff
@@ -655,6 +788,10 @@ void bitfury_do_io(struct thr_info * const master_thr)
 			bitfury->force_reinit = false;
 		}
 	}
+	skip_stat = (skip_stat + 1) & 0x0;
```
luke-jr, line 791: & 0x0 = always 0

Seems skip_stat is completely unused? Mind if I remove it?

asfin, line 791: Ok, you can remove, seems it's deprecated.

luke-jr:
Merged as 5255b8b (and fixed and improved in many commits after that)

PR #309, opened 2013-10-28
Bugfix: OSX: BPMC BF1 would segfault due to some missed copy/pasted bitf...

TrooperT:
...ury code related to polling, which BF1 does not support

commented out code as reminder.


luke-jr:
This is already fixed...

TrooperT:
le derp. >.< my network graph wouldn't update until after i posted

PR #329, opened 2013-11-25
BF2 - Twin Bitfury USB miner driver

aauer1:


driver-bf2.c
```diff
@@ -0,0 +1,467 @@
...
+	       info->id.serial[2],
+	       info->id.serial[3],
+	       info->id.serial[4],
+	       info->id.serial[5],
+	       info->id.serial[6],
+	       info->id.serial[7],
+	       info->id.serial[8],
+	       info->id.serial[9],
+	       info->id.serial[10]
+	                       );
```
luke-jr: serial has only had 4 bytes copied into it here?

driver-bf2.c
```diff
@@ -0,0 +1,467 @@
...
+	if(fd < 0)
+	{
+		return false;
+	}
+
+	char buf[1024];
+	int len;
+
+	serial_read(fd, buf, sizeof(buf));
+	if (1 != write(fd, "I", 1))
```
luke-jr: Another device with a probe-unfriendly protocol? :(

driver-bf2.c
```diff
@@ -0,0 +1,467 @@
...
+	if (1 != write(cgpu->device_fd, buf, 1))
+		return false;
+	mutex_unlock(&cgpu->device_mutex);
+	
+	return true;
+}
+
+//------------------------------------------------------------------------------
+struct device_drv bf2_drv = {
+	.dname = "Twin Bitfury",
```
luke-jr: dname should always match the lowercase filename. so "bf2"

driver-bf2.c
```diff
@@ -0,0 +1,467 @@
...
+	}
+
+	cgpu->device_fd = fd;
+
+	applog(LOG_INFO, "%"PRIpreprv": Opened %s", cgpu->proc_repr, cgpu->device_path);
+
+	info->tx_buffer[0] = 'W';
+	info->tx_buffer[1] = 0x00;
+
+	mutex_init(&cgpu->device_mutex);
```
luke-jr: I don't think you need a mutex? minerloop_async should only call your driver from within a single thread.

driver-bf2.c
aauer1: Ok, we can either rename the file to driver-twinfury.c/h and change dname to "twinfury" or change dname to bf2. What would you prefer?

driver-bf2.c
aauer1: I added a mutex because of the identification function. Is the identification function called within the same thread as minerloop_async? If this is the case, I can remove the mutexes.

driver-bf2.c
aauer1: You are right. 11 bytes should be copied to serial. Will fix this.

driver-bf2.c
aauer1: What do you mean with probe-unfriendly? Any suggestions for better probing is welcome!

driver-bf2.c
luke-jr: Up to you.

driver-bf2.c
luke-jr: Something that doesn't temporary-brick itself when it gets unexpected data (probing for other devices).

driver-bf2.c
luke-jr: Maybe a good idea to use bin2hex here as well.

driver-bf2.c
luke-jr: Ah. You're correct. Usually the way devices handle this is to set a flag (struct cgpu_info has one, though you may need to change an #ifdef for it) and send it at next poll.

aauer1:
I modified my driver for the twinfury:
- Driver file is now called driver-twinfury (dname = "twinfury").
- Flashing the LED changed: flag is set and executed during the next poll cycle (mutex removed)
- serial number is now converted to a hex string with bin2hex utility
- probing the device is the same as before (still bad, but working)

luke-jr:
Will you be changing the protocol to be more tolerant of other probes, or should I merge this?

aauer1:
The device is tolerant if some not recognized commands are sent. The only problem I can currently think of, is when the work command is sent without a work unit. But I think the device should handle this by itself with some kind of timeout. 
So, I would be good with the actual implementation and I would be happy if you could merge this code.

luke-jr:
Want to test it first? My concern is if it will work if you use -S noauto -S all

aauer1:
I will test it. Is this working the same on any operating system (Linux and Windows)?

luke-jr:
Pretty much, yes.

aauer1:
Ok, Luke. I tested the miner with -S noauto -S all and it is working. So, I think the code is now ready to get merged.

luke-jr:
Bah, you broke the Windows build... :p

aauer1:
Oh no. What's the problem on Windows?

luke-jr:
No termios.h or tcflush. Why are you doing tcflush anyway? I don't see a use for it... (it just _discards_ data; see manpage)

aauer1:
oh. you're right. was thinking it will push the data out instead of deleting it. This command has to be removed.

luke-jr:
@aauer1 Please see and test:
- a04dc49 Bugfix: twinfury: Handle flash request for both chips, and avoid using proc uninitialised
- 229097c Bugfix: twinfury: Fix various technicalities
- e6e8c91 Bugfix: twinfury: Remove incorrect tcflush

aauer1:
It is working. Thanks for fixing it.

luke-jr:
np; did you have a way to test on Windows too?

aauer1:
I have a Windows Laptop here and will test it tomorrow.

PR #336, opened 2013-11-30
New / updated instructions for building bfgminer on Windows

nwoolls:
New / updated instructions for building bfgminer on Windows

### DIFF
--- a/windows-build.txt
+++ b/windows-build.txt
@@ -1,50 +1,192 @@
-######################################################################################
-#                                                                                    #
-#          Native WIN32 setup and build instructions (on mingw32/Windows):           #
-#                                                                                    #
-######################################################################################
+*******************
+** Install MinGW **
+*******************
 
-**************************************************************************************
-* Introduction                                                                       *
-**************************************************************************************
-The following instructions have been tested on both Windows 7 and Windows XP.
-Most of what is described below (copying files, downloading files, etc.) can be done
-directly in the MinGW MSYS shell; these instructions do not do so because package
-versions and links change over time. The best way is to use your browser, go to the
-links directly, and see for yourself which versions you want to install.
-Winrar was used to do the extracting of archive files in the making of this guide.
+1.  Go to https://sourceforge.net/downloads/mingw
+2.  Download and execute mingw-get-setup.exe
+3.  Complete the installation wizard leaving default values
 
-**************************************************************************************
-* A tip that might help you along the way                                            *
-**************************************************************************************
-Enable "QuickEdit Mode" in your Command Prompt Window or MinGW Command Prompt
-Window (No need to go into the context menu to choose edit-mark/copy/paste):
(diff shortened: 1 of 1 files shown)


windows-build.txt
```diff
@@ -1,284 +1,269 @@
...
+2.  Type the following:
+
+	cd ~/
+	git clone git://github.com/luke-jr/bfgminer.git
+	cd bfgminer/
+	./autogen.sh
+	CFLAGS="-O2 -msse2 -IC:/MinGW/include/ddk" PKG_CONFIG_PATH=C:/MinGW/msys/1.0/local/lib/pkgconfig ./configure
+	make
+
+* Note: for bfgminer ./configure options, see https://github.com/luke-jr/bfgminer/blob/bfgminer/README
```
luke-jr: README.txt should be included in the same ZIP(s) as this file, no need to go to the web.

windows-build.txt
```diff
@@ -1,284 +1,269 @@
...
+**********************
+
+1.  Return to the MinGW Shell
+2.  Type the following:
+
+	cd ~/
+	git clone git://github.com/luke-jr/bfgminer.git
+	cd bfgminer/
+	./autogen.sh
+	CFLAGS="-O2 -msse2 -IC:/MinGW/include/ddk" PKG_CONFIG_PATH=C:/MinGW/msys/1.0/local/lib/pkgconfig ./configure
```
luke-jr: Variable assignments should be to the right of the configure command:

```
./configure CFLAGS='-O2 -msse2 -IC:/MinGW/include/ddk' PKG_CONFIG_PATH='C:/MinGW/msys/1.0/local/lib/pkgconfig'
```

Also, I think CFLAGS does not need to be overridden/specified?

windows-build.txt
```diff
@@ -1,284 +1,269 @@
...
+	libiconv-2.dll
+
+C:\MinGW\msys\1.0\local\bin  
+
+	libevent-2-0-5.dll
+
+C:\MinGW\msys\1.0\home\USER\bfgminer\libblkmaker\.libs  
+
+	libblkmaker-0.1-0.dll
+	libblkmaker_jansson-0.1-0.dll
```
luke-jr, line 242: Missing newline at  the end.

windows-build.txt
nwoolls: I'll fix up the formatting. The -I is needed in order to find usbioctl.h. The others were carried over from the previous instructions.

luke-jr: It shouldn't be needed to find it... configure checks for the subdirectory.

nwoolls: I've needed it on two separate systems now, verified. I don't know why configure isn't finding it. Maybe because I install MinGW to my D drive? Either way it's definitely needed on both my 32-bit and 64-bit Windows VMs.

I am cleaning up the instructions now and verifying them again on a clean system. I'll let you know once thats completed.

PR #348, opened 2013-12-12
Twinfury voltage scaling

aauer1:
Hi Luke,
I would like to add a voltage scaling feature for the twinfury miner. Here are my code changes. I would really appriciate if you could merge it into your bfgminer repository. 

Thank you!

### DIFF
--- a/driver-twinfury.c
+++ b/driver-twinfury.c
@@ -114,34 +114,6 @@ static bool twinfury_detect_custom(const char *devpath, struct device_drv *api, 
 	       info->id.version, info->id.product,
 	       info->id.serial);
 
-	if(info->id.version == 2)
-	{
-		char buf[8] = "V\x00\x00";
-		if(twinfury_send_command(fd, buf, 3))
-		{
-			if(8 == twinfury_wait_response(fd, buf, 8))
-			{
-				info->voltage  =  (buf[4] & 0xFF);
-				info->voltage |=  (buf[5] << 8);
-
-				applog(LOG_DEBUG, "%s: Voltage: %dmV", twinfury_drv.dname, info->voltage);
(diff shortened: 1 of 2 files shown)


driver-twinfury.c
```diff
@@ -114,6 +114,33 @@ static bool twinfury_detect_custom(const char *devpath, struct device_drv *api,
...
+
+				applog(LOG_DEBUG, "Voltage: %d", info->voltage);
+				if(info->voltage < 800 || info->voltage > 950)
+				{
+					info->voltage = 0;
+				}
+			}
+			else
+			{
+				applog(LOG_DEBUG, "FAILED TO GET VOLTAGE");
```
luke-jr: Errors should be errors unless they're normal/expected.. and not all uppercase. Probably good to mention which device (dev_repr) too... :)

aauer1: You are right. Changed it to LOG_ERR, added the driver name to the logging output. text is now upper- and lowercase :)

driver-twinfury.c
```diff
@@ -114,6 +114,34 @@ static bool twinfury_detect_custom(const char *devpath, struct device_drv *api,
 	       info->id.version, info->id.product,
 	       info->id.serial);
 
+	if(info->id.version == 2)
```
luke-jr: It is probably better to re-read voltage when opening device later, to avoid a race and allow it to be performed when things are parallel :)

aauer1: So you mean that I should do this better in the thread_init function?!

luke-jr: Yes, that is probably a good place.

aauer1:
Moved the twinfury voltage reading to the thread init function.

luke-jr:
Thanks, merged.

PR #353, opened 2013-12-26
experimental arm NEON cpu algo modeled after sse2 4way algo

TrooperT:
Notes: several hacks were needed to get her running.

must pass CFLAGS="-g -O2 -mfpu=neon" ./configure for it to compile. or directly modify bfgminer_CPPFLAGS in  generated Makefile

had to disable default c algo to get NEON enabled :/

reported hash rate on BBB ~305kHash/sec

did not run long enough to find nonce (validation that code is correct) but it was directly modeled after sha256_4Way.c only modifying the sse2 intrinsics to NEON 
screen shot 2013-12-26 at 1 11 41 am

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -138,7 +138,8 @@ bfgminer_SOURCES	+= \
 		  sha256_generic.c sha256_via.c	\
 		  sha256_cryptopp.c sha256_sse2_amd64.c		\
 		  sha256_sse4_amd64.c 	\
-		  sha256_altivec_4way.c
+		  sha256_altivec_4way.c \
+		  sha256_NEON_4way.c
 
 # the CPU portion extracted from original main.c
 bfgminer_SOURCES += driver-cpu.h driver-cpu.c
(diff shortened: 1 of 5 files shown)


luke-jr:
You can test it with

```
./bfgminer -o http://luke.dashjr.org/tmp/code/earlyshare.json -O x
```

It should find a (Rejected) share fairly fast.

TrooperT:
well i will have to take a look over the code. even though it says its hashing, i tried all 4 of the earlyshare**.json i found and set scantime and expiry to 300 and found nothing

TrooperT:
le sigh

TrooperT:
ran ./bfgminer -o http://luke.dashjr.org/tmp/code/earlyshare.json -O x 

with an AMU to be sure and the nonce is at 38fc34ec, which while early would take 300Khash ~55minutes to reach

would need test target with nonce in first ~18million( 112A880) nonce space to get it in just under 1 minute of scantime. if anyone has one let me know

luke-jr:
38fc34ec is the hash, not the nonce. The nonce is b3380000 (or 0x38b3 little-endian).

TrooperT:
We'll i def has a problem then ~:[

luke-jr:
Ping me when you're ready to reopen...

PR #355, opened 2013-12-27
Allow overriding the maximum frequency used by the dynclock logic

nwoolls:
Can now use --set-device x6500:maxclock=210
Prevents spending time on frequencies that only produce HW errors

### DIFF
--- a/driver-x6500.c
+++ b/driver-x6500.c
@@ -738,13 +738,23 @@ void x6500_user_set_clock(struct cgpu_info *cgpu, const int val)
 }
 
 static
+void x6500_user_set_max_clock(struct cgpu_info *cgpu, const int val)
+{
+	struct thr_info * const thr = cgpu->thr[0];
+	struct x6500_fpga_data *fpga = thr->cgpu_data;
+	const int multiplier = val / 2;
+	fpga->freqMaxMaxM =
+	fpga->dclk.freqMaxM = multiplier;
+}
+
+static
 char *x6500_set_device(struct cgpu_info *cgpu, char *option, char *setting, char *replybuf)
 {
 	int val;
 	
 	if (strcasecmp(option, "help") == 0) {
-		sprintf(replybuf, "clock: range %d-%d and a multiple of 2",
-		        X6500_MINIMUM_CLOCK, X6500_MAXIMUM_CLOCK);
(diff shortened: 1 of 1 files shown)


driver-x6500.c
```diff
@@ -765,6 +775,29 @@ char *x6500_set_device(struct cgpu_info *cgpu, char *option, char *setting, char
...
+	if (strcasecmp(option, "maxclock") == 0) {
+		if (!setting || !*setting) {
+			sprintf(replybuf, "missing maxclock setting");
+			return replybuf;
+		}
+		
+		val = atoi(setting);
+		if (val < X6500_MINIMUM_CLOCK || (val & 1) != 0) {
+			sprintf(replybuf, "invalid clock: '%s' valid range %d+ and a multiple of 2",
+			        setting, X6500_MINIMUM_CLOCK);
```
luke-jr: This should enforce <= X6500_MAXIMUM_CLOCK too.

driver-x6500.c
```diff
@@ -765,6 +775,29 @@ char *x6500_set_device(struct cgpu_info *cgpu, char *option, char *setting, char
...
+			return replybuf;
+		}
+		
+		x6500_user_set_max_clock(cgpu, val);
+                
+                applog(LOG_NOTICE, "%"PRIpreprv": Maximum frequency reset to %u MHz",
+                        cgpu->proc_repr,
+                        val
+                );
```
luke-jr: Indentation is off here. Please indent with tabs, and align with spaces. See driver-ztex.c:92 for a multiline applog example:

```
        applog(LOG_DEBUG, "%s: Matched \"%s\" serial \"%s\", but lowlevel driver is not usb!",
               __func__, product, serial);
```

luke-jr:
Please use for commit message format: ["Bugfix: "][dname-or-function-or-generalarea ":"]description

driver-x6500.c
nwoolls: It could, but then the user couldn't pick a maxclock value that is greater than 250. In other words this could only be used to lower the maxclock, not raise it. That's fine for my personal use-case.

Should this only be for lowering the maxclock?

nwoolls:
Updated with your comments in mind

PR #356, opened 2013-12-27
Minor - fix an extra space when a frequency is raised

nwoolls:
Minor - fix an extra space when a frequency is raised


luke-jr:
This is intentional so it lines up with the frequency decrease.

nwoolls:
10-4

Issue #338, opened 2013-12-02
Bus Error 10 on Mac with Butterfly BitForce ASIC with bfgminer 3.8 and 3.7


gthiruva:
Looks like something with the probing functionality broke after the 3.6 release (at least with the Butterfly 60 GH/s BitForce SC).

When probing devices with the 3.7/3.8 tags bfgminer dies right after locking the USB device (presumable when trying to open and access the device):

```
$ ./bfgminer -D -T --verbose -S all
 [2013-12-02 12:33:00] setrlimit: Changed soft fd limit from 256 to 1024 (FD_SETSIZE=1024; hard limit=unlimited)                    
 [2013-12-02 12:33:00] Started bfgminer 3.8.0                    
 [2013-12-02 12:33:00] lowlevel_scan: Found vcom device at dev_t:1d000012 (path=/dev/cu.usbserial-000011FD, vid=0000, pid=0000, manuf=(null), prod=(null), serial=(null))                    
 [2013-12-02 12:33:00] lowlevel_scan: Found vcom device at dev_t:03000012 (path=/dev/cu.Bluetooth-Modem, vid=0000, pid=0000, manuf=(null), prod=(null), serial=(null))                    
 [2013-12-02 12:33:00] lowlevel_scan: Found vcom device at dev_t:01000012 (path=/dev/cu.Bluetooth-PDA-Sync, vid=0000, pid=0000, manuf=(null), prod=(null), serial=(null))                    
 [2013-12-02 12:33:00] Icarus Detect: Attempting to open /dev/cu.Bluetooth-Modem                    
 [2013-12-02 12:33:00] Icarus Detect: Attempting to open /dev/cu.Bluetooth-PDA-Sync                    
 [2013-12-02 12:33:00] Open /dev/cu.Bluetooth-Modem failed: Resource busy                    
 [2013-12-02 12:33:00] Icarus Detect: Attempting to open /dev/cu.usbserial-000011FD                    
 [2013-12-02 12:33:00] Icarus Detect: Failed to open /dev/cu.Bluetooth-Modem                    
 [2013-12-02 12:33:00] Icarus Detect: Attempting to open /dev/cu.Bluetooth-Modem                    
 [2013-12-02 12:33:00] Open /dev/cu.Bluetooth-Modem failed: Resource busy                    
 [2013-12-02 12:33:00] Icarus Detect: Failed to open /dev/cu.Bluetooth-Modem                    
[...]

PR #358, opened 2013-12-27
Bugfix: don’t attempt to probe Bluetooth devices when scanning hardware

nwoolls:
Causes bus errors and segfaults on OS X


nwoolls:
Fixes Issue #338 along with a segfault and allows Twinfury to be recognized on OS X

PR #359, opened 2013-12-28
Bugfix: Do not allocate spi_port on the Stack - EXC_BAD_ACCESS on OS X

nwoolls:


luke-jr:
Please make the function names unique (eg, dname_fname). I realise it's mostly safe since they're static, but just in case...

nwoolls:
Thanks good call still not used to C - this is going to take a while!

nwoolls:
Done.

nwoolls:
Comments updated

PR #364, opened 2014-01-04
Bugfix: HID API not properly detected on Mac OS X

nwoolls:
Mach-O uses dylibs for shared libraries: http://www.finkproject.org/doc/porting/porting.en.html#shared


luke-jr:
Please use #elif defined(...)

PR #368, opened 2014-01-06
antminer: Initial support for the Bitmain AntMiner U1 ASIC

nwoolls:


driver-icarus.c
```diff
@@ -1110,10 +1111,15 @@ static int64_t icarus_scanhash(struct thr_info *thr, struct work *work,
...
-	hash_count++;
-	hash_count *= info->fpga_count;
+	
+	if (info->do_estimate_hashrate) {
+		hash_count = (nonce & info->nonce_mask);
+		hash_count++;
+		hash_count *= info->fpga_count;
+	} else {
+		//Bitmain AntMiner U1
+		hash_count = 0xffffffff;
```
luke-jr: This doesn't make sense...

driver-icarus.c
```diff
@@ -734,7 +729,7 @@ static bool icarus_init(struct thr_info *thr)
 			"BFG\0\x64\x61\x01\x1a\xc9\x06\xa9\x51\xfb\x9b\x3c\x73";
 		
 		icarus_write(fd, pkt, sizeof(pkt));
-		if (ICA_GETS_OK == icarus_gets((unsigned char*)&res, fd, &tv_finish, NULL, info->read_count))
+		if (ICA_GETS_OK == icarus_gets((unsigned char*)&res, fd, &tv_finish, NULL, info->read_count, info->read_size))
```
luke-jr: Buffer overflow: res is 32-bit

driver-icarus.c
```diff
@@ -882,7 +877,7 @@ void handle_identify(struct thr_info * const thr, int ret, const bool was_first_
 				break;
 			
 			// Try to get more nonces (ignoring work restart)
-			ret = icarus_gets((void *)&nonce, fd, &tv_now, NULL, (info->fullnonce - delapsed) * 10);
+			ret = icarus_gets((void *)&nonce, fd, &tv_now, NULL, (info->fullnonce - delapsed) * 10, info->read_size);
```
luke-jr: Buffer overlow

driver-icarus.c
```diff
@@ -979,7 +974,7 @@ static int64_t icarus_scanhash(struct thr_info *thr, struct work *work,
 			read_count = info->read_count;
 keepwaiting:
 			/* Icarus will return 4 bytes (ICARUS_READ_SIZE) nonces or nothing */
-			ret = icarus_gets((void*)&nonce, fd, &state->tv_workfinish, thr, read_count);
+			ret = icarus_gets((void*)&nonce, fd, &state->tv_workfinish, thr, read_count, info->read_size);
```
luke-jr: Buffer overlow

driver-antminer.c
```diff
@@ -0,0 +1,89 @@
...
+//	struct icarus_state *state = thr->cgpu_data;
+//	state->identify = true;
+//	return true;
+//}
+
+static void antminer_drv_init()
+{
+	antminer_drv = icarus_drv;
+	antminer_drv.dname = "antminer";
+	antminer_drv.name = "ANT";
```
luke-jr, line 231: AMU

driver-antminer.c
```diff
@@ -0,0 +1,89 @@
...
+	antminer_drv.lowl_probe = antminer_lowl_probe;
+//	antminer_drv.identify_device = antminer_identify;
+	++antminer_drv.probe_priority;
+}
+
+struct device_drv antminer_drv = {
+	.drv_init = antminer_drv_init,
+};
+
+struct device_drv antminer_drv_emerald = {
```
luke-jr: Huh?

driver-antminer.c
nwoolls, line 231: Done to avoid any confusion, talking points, etc. whatsoever with cgminer, which uses AMU to stand for ASICMINER Unit (Block Erupter). On the one hand, worrying about the terminology in another tool seems unnecessary. But, given the history and tendency for drama, it seemed a wise concession.

driver-antminer.c
nwoolls: Forgot that one - I'll take care of it.

driver-icarus.c
nwoolls: I agree, but unfortunately I don't know enough of the maths involved. The weighed accepted shares along with the u: value in bfgminer both correspond to the values that are displayed by using this code. From talking to Wizkid on IRC, he indicated that this code would be valid if the device scans the full nonce. You indicated no Icarus device does - but maybe this one does?

driver-icarus.c
nwoolls: Changed from uint32_t to uint64_t.

driver-icarus.c
nwoolls: Changed from uint32_t to uint64_t.

driver-icarus.c
nwoolls: Changed from uint32_t to uint64_t.

nwoolls: I'm going to revisit this. While it avoids issues in practice with the AntMiner and a read_size of 5, I think that in theory the value could be larger which could still lead to an overflow.

typoworx-de:
Already tried BfgMiner (current GIT code) to be used with Antminer/U1 (USB Version). They're working but the speed does not match. After starting Bfgminer they're currently running up and shortly they peak at 1,5-1,6 GH/s, but then the speed is falling down to 300-500 MH/s.

Is this because the current implementation is not complete yet?

What is about the overclocking/speed parameters currently "experimentally" existing in CgMiner? Didn't figure out how to do that. Or isn't it integrated yet?

nwoolls:
Make sure you are using the AntMiner driver and not the Icarus driver. You must use the -S antminer argument on the command line.

I've tested the code with a single AntMiner U1 on Windows, OS X, and Raspberry Pi and on all platforms it hashes at 1.6 Gh/s.

typoworx-de:
Thanks for the hint on the -S Parameter. I quickly checked it out and bfgminer (version 3.9.0) is running with giving "-S antminer" to it as argument (exactly as you wrote above).

But the Antminer-Devices are still shown as "ICA" and still having performance lack. Do I have something missing or a wrong version of bfgminer?

I already tried to compile cgminer that was told to be "compatible" in some patched versions and finally found a version running with good performance.

Never the less I would prefer to use bfgminer ;-)

Is overclocking feature also possible with bfgminer? Which parameter do I have to use to switch clock-mode? In cgminer I use "--bmsc options 115200:20" for the UART and "--bmsc-freq [integer-id]" for the clock-mode. Is there something like that for cgminer?

nwoolls:
There is no overclocking yet and the --bmsc options 115200:20 argument is not required the way it is with cgminer. The AntMiner U1 "just works" with bfgminer. If it is showing ICA that means it is not using the AntMiner driver and they will run slow and with about 36% HW errors.

If you compiled this yourself, make sure you checked out the feature/antminer-u1-support branch. Check for the driver-antminer.c file being present. I also linked to precompiled Windows EXEs on the forums.

Scopiat:
how do we get notified when the overclocking feature gets implemented?

nwoolls:
@Scopiat already implemented, you're commenting on an old closed Pullreq. New one is here:
https://github.com/luke-jr/bfgminer/pull/373

bfgminer -S antminer:all --set-device antminer:freq=ABCD

typoworx-de:
Thank you @nwoolls, now it works like a charm after re-cloning the current git version.

ghost:
Thanks for this thread, with its help, I have bfgminer working great with (2) AMUs under Gentoo Linux (emerged bfgminer to get all the support libraries and tested using the GPU, then did the git version for antminer support).

Currently burning in those two AMUs and noticing one small thing after a few hours:  AMU 0 is running around 1.5Gh/s, where AMU 1 is running at the expected 1.6Gh/s.  They both bounce around a bit, but that 0.1 Gh/s difference seems to be fairly consistent.

Starting with:  ./bfgminer -S antminer:all --set-device antminer.freq=0781 (and all the pool options of course)

ghost:
Having a problem getting the --set-device to work.

Grabbed code via:  git clone https://github.com/luke-jr/bfgminer
Did a unmodified ./autogen.sh & ./configure

Passing bfgminer the string "-S antminer:all --set-device antminer:freq=0A81"

and getting the error during startup of:  AMU 0: Applying rule antminer:freq=0A81: Unknown option: freq

I'm missing whatever I'm doing wrong.  Help please!

(of course, I'm seeing that message on the other AMU's as well, and they are otherwise running fine, just at standard speeds)

ghost:
Ah... looked in the antminer code.  Apparently the option is now "clock" and it requires a leading x.  This worked:

"-S antminer:all --set-device antminer:clock=x0981"

luke-jr:
This is not a discussion forum. It is a closed pull request.

ghost:
My apologies.

Is there an appropriate discussion forum for this?

Cassey

luke-jr:
https://bitcointalk.org/?topic=168174

PR #369, opened 2014-01-06
minorfix: remove extra semicolon

isghe:
very minor fix


luke-jr:
What's wrong with the extra semicolons?

isghe:
Just in case: they are C legal empty statements; depending on the implementations, the C compiler can decide to add or not to the assembly result as, anyway expensive, NOP assembly code; "typically", no instructions whatsoever would be generated.

luke-jr:
Merged, thanks.

PR #373, opened 2014-01-09
antminer: Initial support for the Bitmain AntMiner U1 ASIC

nwoolls:
Includes support for identifying the U1 separately from Icarus and Block Erupter
Also includes overclocking via --set-device antminer:freq=HEX

### DIFF
--- a/driver-antminer.c
+++ b/driver-antminer.c
@@ -1,0 +1,188 @@
+/*
+ * Copyright 2013 Luke Dashjr
+ * Copyright 2013 Nate Woolls
+ * Copyright 2013 Lingchao Xu
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 3 of the License, or (at your option)
+ * any later version.  See COPYING for more details.
+ */
+
+#include "config.h"
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+
+#include "miner.h"
+#include "icarus-common.h"
+#include "lowlevel.h"
+#include "lowl-vcom.h"
+#include "deviceapi.h"
+#include "logging.h"
+#include "util.h"
+
+#define ANTMINER_IO_SPEED 115200
+#define ANTMINER_HASH_TIME 0.0000000004761
+
+#define ANTMINER_STATUS_LEN 5
+
+BFG_REGISTER_DRIVER(antminer_drv)
+
+static
+bool antminer_detect_one(const char *devpath)
+{
+	struct device_drv *drv = &antminer_drv;
+	
+	struct ICARUS_INFO *info = calloc(1, sizeof(struct ICARUS_INFO));
+	if (unlikely(!info))
+		quit(1, "Failed to malloc ICARUS_INFO");
+	
+	*info = (struct ICARUS_INFO){
+		.baud = ANTMINER_IO_SPEED,
+		.Hs = ANTMINER_HASH_TIME,
+		.timing_mode = MODE_DEFAULT,
+		.read_size = 5,
+	};
+	
+	if (!icarus_detect_custom(devpath, drv, info))
+	{
+		free(info);
+		return false;
+	}
+	
+	info->read_count = 15;
+	
+	return true;
+}
+
+static
+bool antminer_lowl_probe(const struct lowlevel_device_info * const info)
+{
+	return vcom_lowl_probe_wrapper(info, antminer_detect_one);
+}
+
+static
+char *antminer_get_clock(struct cgpu_info *cgpu, char *replybuf)
+{
+	uint8_t rdreg_buf[4] = {0};
+	unsigned char rebuf[ANTMINER_STATUS_LEN] = {0};
+	
+	struct timeval tv_now;
+	
+	rdreg_buf[0] = 4;
+	rdreg_buf[0] |= 0x80;
+	rdreg_buf[1] = 0;    //16-23
+	rdreg_buf[2] = 0x04; // 8-15
+	rdreg_buf[3] = crc5usb(rdreg_buf, 27);
+	
+	applog(LOG_DEBUG, "%"PRIpreprv": Get clock: %02x%02x%02x%02x", cgpu->proc_repr, rdreg_buf[0], rdreg_buf[1], rdreg_buf[2], rdreg_buf[3]);
+	
+	timer_set_now(&tv_now);
+	int err = icarus_write(cgpu->device_fd, rdreg_buf, sizeof(rdreg_buf));
+	
+	if (err != 0)
+	{
+		sprintf(replybuf, "invalid send get clock: comms error (err=%d)", err);
+		return replybuf;
+	}
+	
+	applog(LOG_DEBUG, "%"PRIpreprv": Get clock: OK", cgpu->proc_repr);
+	
+	memset(rebuf, 0, sizeof(rebuf));
+	err = icarus_gets(rebuf, cgpu->device_fd, &tv_now, NULL, 10, ANTMINER_STATUS_LEN);
+	
+	// Timeout is ok - checking specifically for an error here
+	if (err == ICA_GETS_ERROR)
+	{
+		sprintf(replybuf, "invalid recv get clock: comms error (err=%d)", err);
+		return replybuf;
+	}
+		
+	applog(LOG_DEBUG, "%"PRIpreprv": Get clock: %02x%02x%02x%02x%02x", cgpu->proc_repr, rebuf[0], rebuf[1], rebuf[2], rebuf[3], rebuf[4]);
+	
+	return NULL;
+}
+
+static
+char *antminer_set_clock(struct cgpu_info *cgpu, char *setting, char *replybuf)
+{
+	if (!setting || !*setting)
+		return "missing clock setting";
+	
+	// For now we only allow hex values that use BITMAINtech's lookup table
+	// This means values should be prefixed with an x so that later we can
+	// accept and distinguish decimal values
+	if (setting[0] != 'x')
+	{
+		sprintf(replybuf, "invalid clock: '%s' data must be prefixed with an x", setting);
+		return replybuf;
+	}
+	
+	//remove leading character
+	char *hex_setting = setting + 1;
+
+	uint8_t reg_data[4] = {0};
+	
+	if (!hex2bin(reg_data, hex_setting, strlen(hex_setting) / 2))
+	{
+		sprintf(replybuf, "invalid clock: '%s' data must be a hexidecimal value", hex_setting);
+		return replybuf;
+	}
+	
+	uint8_t cmd_buf[4] = {0};
+	
+	cmd_buf[0] = 2;
+	cmd_buf[0] |= 0x80;
+	cmd_buf[1] = reg_data[0]; //16-23
+	cmd_buf[2] = reg_data[1]; // 8-15
+	cmd_buf[3] = crc5usb(cmd_buf, 27);
+	
+	applog(LOG_DEBUG, "%"PRIpreprv": Set clock: %02x%02x%02x%02x", cgpu->proc_repr, cmd_buf[0], cmd_buf[1], cmd_buf[2], cmd_buf[3]);
+	
+	int err = icarus_write(cgpu->device_fd, cmd_buf, sizeof(cmd_buf));
+		
+	if (err != 0)
+	{
+		sprintf(replybuf, "invalid send clock: '%s' comms error (err=%d)", setting, err);
+		return replybuf;
+	}
+	
+	applog(LOG_DEBUG, "%"PRIpreprv": Set clock: OK", cgpu->proc_repr);
+	
+	// This is confirmed required in order for the clock change to "take"
+	cgsleep_ms(500);
+		
+	return antminer_get_clock(cgpu, replybuf);
+}
+
+static
+char *antminer_set_device(struct cgpu_info *cgpu, char *option, char *setting, char *replybuf)
+{		
+	if (strcasecmp(option, "clock") == 0)
+	{
+		return antminer_set_clock(cgpu, setting, replybuf);
+	}
+
+	sprintf(replybuf, "Unknown option: %s", option);
+	return replybuf;
+}
+
+static
+void antminer_drv_init()
+{
+	antminer_drv = icarus_drv;
+	antminer_drv.dname = "antminer";
+	antminer_drv.name = "AMU";
+	antminer_drv.lowl_probe = antminer_lowl_probe;
+	antminer_drv.set_device = antminer_set_device,
+	++antminer_drv.probe_priority;
+}
+
+struct device_drv antminer_drv = {
+	.drv_init = antminer_drv_init,
+};
--- a/driver-cairnsmore.c
+++ b/driver-cairnsmore.c
@@ -92,7 +92,7 @@ bool cairnsmore_supports_dynclock(int fd)
 			.work_restart = false,
 			.work_restart_notifier = {-1, -1},
 		};
-		icarus_gets((unsigned char*)&nonce, fd, &tv_finish, &dummy, 1);
+		icarus_gets((unsigned char*)&nonce, fd, &tv_finish, &dummy, 1, ICARUS_DEFAULT_READ_SIZE);
 	}
 	applog(LOG_DEBUG, "Cairnsmore dynclock detection... Got %08x", nonce);
 	switch (nonce) {
--- a/driver-icarus.c
+++ b/driver-icarus.c
@@ -65,17 +65,15 @@
 // The serial I/O speed - Linux uses a define 'B115200' in bits/termios.h
 #define ICARUS_IO_SPEED 115200
 
-// The size of a successful nonce read
-#define ICARUS_READ_SIZE 4
+// The number of bytes in a nonce (always 4)
+// This is NOT the read-size for the Icarus driver
+// That is defined in ICARUS_INFO->read_size
+#define ICARUS_NONCE_SIZE 4
 
-// Ensure the sizes are correct for the Serial read
-#if (ICARUS_READ_SIZE != 4)
-#error ICARUS_READ_SIZE must be 4
-#endif
 #define ASSERT1(condition) __maybe_unused static char sizeof_uint32_t_must_be_4[(condition)?1:-1]
 ASSERT1(sizeof(uint32_t) == 4);
 
-#define ICARUS_READ_TIME(baud) ((double)ICARUS_READ_SIZE * (double)8.0 / (double)(baud))
+#define ICARUS_READ_TIME(baud, read_size) ((double)read_size * (double)8.0 / (double)(baud))
 
 // Defined in deciseconds
 // There's no need to have this bigger, since the overhead/latency of extra work
@@ -184,18 +182,13 @@ static void rev(unsigned char *s, size_t l)
 #define icarus_open2(devpath, baud, purge)  serial_open(devpath, baud, ICARUS_READ_FAULT_DECISECONDS, purge)
 #define icarus_open(devpath, baud)  icarus_open2(devpath, baud, false)
 
-#define ICA_GETS_ERROR -1
-#define ICA_GETS_OK 0
-#define ICA_GETS_RESTART 1
-#define ICA_GETS_TIMEOUT 2
-
-int icarus_gets(unsigned char *buf, int fd, struct timeval *tv_finish, struct thr_info *thr, int read_count)
+int icarus_gets(unsigned char *buf, int fd, struct timeval *tv_finish, struct thr_info *thr, int read_count, int read_size)
 {
 	ssize_t ret = 0;
 	int rc = 0;
 	int epollfd = -1;
 	int epoll_timeout = ICARUS_READ_FAULT_DECISECONDS * 100;
-	int read_amount = ICARUS_READ_SIZE;
+	int read_amount = read_size;
 	bool first = true;
 
 #ifdef HAVE_EPOLL
@@ -282,7 +275,7 @@ int icarus_gets(unsigned char *buf, int fd, struct timeval *tv_finish, struct th
 	}
 }
 
-static int icarus_write(int fd, const void *buf, size_t bufLen)
+int icarus_write(int fd, const void *buf, size_t bufLen)
 {
 	size_t ret;
 
@@ -557,6 +550,27 @@ static void get_options(int this_option_offset, struct ICARUS_INFO *info)
 	}
 }
 
+// Number of bytes remaining after reading a nonce from Icarus
+int icarus_excess_nonce_size(int fd, struct ICARUS_INFO *info)
+{
+	// How big a buffer?
+	int excess_size = info->read_size - ICARUS_NONCE_SIZE;
+
+	// Try to read one more to ensure the device doesn't return
+	// more than we want for this driver
+	excess_size++;
+
+	unsigned char excess_bin[excess_size];
+	// Read excess_size from Icarus
+	struct timeval tv_now;
+	timer_set_now(&tv_now);
+	//icarus_gets(excess_bin, fd, &tv_now, NULL, 1, excess_size);
+	int bytes_read = read(fd, excess_bin, excess_size);
+	// Number of bytes that were still available
+
+	return bytes_read;
+}
+
 bool icarus_detect_custom(const char *devpath, struct device_drv *api, struct ICARUS_INFO *info)
 {
 	int this_option_offset = ++option_offset;
@@ -582,7 +596,7 @@ bool icarus_detect_custom(const char *devpath, struct device_drv *api, struct IC
 
 	const char golden_nonce[] = "000187a2";
 
-	unsigned char ob_bin[64], nonce_bin[ICARUS_READ_SIZE];
+	unsigned char ob_bin[64], nonce_bin[ICARUS_NONCE_SIZE];
 	char nonce_hex[(sizeof(nonce_bin) * 2) + 1];
 
 	get_options(this_option_offset, info);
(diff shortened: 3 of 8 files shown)


1am3r:
Just compiled and installed your version. It seems to run fine, only the 5s and the all-time average hashrate seem to be quite off. The submitted shares based average seems to be right and reflects the hashrate my pool shows me. Is this a known bug or did i do something wrong?

EDIT: Also after running them for ~3 hours i have no hardware errors on any of the 10 sticks. They definitely had some when running under cgminer, so either your driver is much better or the HW-Errors do not get reported :)

nwoolls:
Off compared to what? Try running it through a Proxy and checking the hashrate shown there compared to the hashrate shown in bfgminer. The U1 hashrate jumps all over the place and the cgminer build from BITMAINtech hides this with hard-coded hashrate values.

1am3r:
Off like 0.96/ 0.84/ 2.00Gh/s as an example. The first and second value never go over 1.2Gh/s.
I'll run them through a proxy later and report back.

nwoolls:
Thanks for the details. If you can take note of what the values are after 5 minutes, then 10, then 15, and then maybe try on more than one OS or machine. I've tested on three difference machines & OSs and checked through a proxy and after mining for a few minutes things have looked right on my end.

Also note that this only affects the hashrate displayed, not the actual hashing or accepted shares.

1am3r:
Yeah its just the displayed hashrate, the submitted rate is fine. 
I'm currently running my ants at 2Gh/s on a cubieboard2 under debian through a local proxy and i'm getting the following values:

-after 5min:
proxy: 1            | 22.13/19.76/19.76Gh/s | A:123 R:0+0(none) HW:0/none
direct: 10           |  8.01/ 8.41/19.80Gh/s | A:2029 R:0+0(none) HW:0/none

-after 10min:
proxy: 1            | 21.49/20.36/20.20Gh/s | A:253 R:2+0(.80%) HW:0/none
direct: 10           |  9.52/ 8.55/20.33Gh/s | A:3046 R:31+0(1.0%) HW:1/.03%

-after 15min:
proxy: 1            | 20.24/20.58/20.48Gh/s | A:323 R:2+0(.58%) HW:0/none
direct: 10           |  7.76/ 8.60/20.58Gh/s | A:4335 R:31+0(.71%) HW:1/.02%

So with mining with difficulty 1 (had to use the bfgminer http proxy as the stratum one wouldnt work) a HW-error got reported. I'll let them run directly to the pool tonight to see if HW-errors get reported with higher diffs, too.

Ill try to compile it for windows now and see how it works there.

nwoolls:
Interesting - I haven't had any other reports of the hashrate being off (so far), but I don't doubt there could be something off either. Basically the hashrate is estimated using a constant that is specific to each device. In the case of cgminer's AntMiner support the hashrate is not estimated at all.

Does adjusting the overclock frequency up or down provide more reliable results?

1am3r:
I'm still in the in the process of compiling it for windows, so i haven't tested it there. 

But I just changed the overclock frequency back to 1.6Hh/s and to 1.8Gh/s:

1.6Gh/s:
ANT 0:       |  0.82/ 0.67/ 1.57Gh/s | A:  6 R:0+0(none) HW: 0/none

1.8Gh/s:
ANT 0:       |  0.90/ 0.78/ 1.92Gh/s | A: 2 R:0+0(none) HW: 0/none

2.0Gh/s:
ANT 0:       |  0.85/ 0.90/ 2.31Gh/s | A:  7 R:0+0(none) HW: 0/none

EDIT: Just tested it under windows. The hasrate gets reported correctly there. So Im not sure what's up...

jimihendrix857:
I was doing some research on the antminer U1 sticks being used on a USB hub and came across this same issue where Im running bfgminer and its mapping as a ICA device and running with 37+% hardware issues....Im guessing cause there is no AMU support in bfgminer 3.9.0. I was hoping someone could compile a windows install of bfgminer that will support Antminers, since i have no background in programming. 
Thanks in advance

nwoolls:
@jimihendrix857 Unofficial Windows binaries with AntMiner U1 support are available from https://www.dropbox.com/s/z29x4n6ng510joi/bfgminer-npw-antminer.7z

nwoolls:
@1am3r interesting - is there any way you'd be willing to install Debian on an x86 box and test there? If it works there then it would seem to me it's something to do with running it on the Cubieboard2. I was actually considering getting one of those myself but was advised to go with a Beaglebone Black instead.

If it's an issue with the Cubieboard2 specifically maybe you could allow me to SSH in and take a look?

driver-antminer.c
```diff
@@ -0,0 +1,171 @@
...
+#include "lowl-vcom.h"
+
+#define ANTMINER_IO_SPEED 115200
+#define ANTMINER_HASH_TIME 0.0000000004761
+
+#define ANTMINER_STATUS_LEN 8192
+
+BFG_REGISTER_DRIVER(antminer_drv)
+
+unsigned char antminer_crc5(unsigned char *ptr, unsigned char len)
```
luke-jr: Is this a standard CRC?

driver-antminer.c
```diff
@@ -0,0 +1,171 @@
...
+{
+    unsigned char i, j, k;
+    unsigned char crc = 0x1f;
```
luke-jr: Prefer uint8_t for this.

driver-antminer.c
```diff
@@ -0,0 +1,171 @@
...
+    unsigned char crcout[5] = {1, 1, 1, 1, 1};
+    unsigned char din = 0;
+
+    j = 0x80;
+    k = 0;
+    for (i = 0; i < len; i++)
+    {
+    	if (*ptr & j) {
+    		din = 1;
+    	} else {
```
luke-jr: No need for braces here.

driver-antminer.c
```diff
@@ -0,0 +1,171 @@
...
+	return vcom_lowl_probe_wrapper(info, antminer_detect_one);
+}
+
+static
+char *antminer_set_frequency(struct cgpu_info *cgpu, char *setting, char *replybuf)
+{
+	unsigned char reg_data[4] = {0};
+	unsigned char cmd_buf[4] = {0};
+	
+	if (!hex2bin(reg_data, setting, strlen(setting) / 2)) {
```
luke-jr: Why is this hex? Should be just a number of MHz...?

driver-cairnsmore.c
```diff
@@ -78,13 +78,16 @@ static bool cairnsmore_send_cmd(int fd, uint8_t cmd, uint8_t data, bool probe)
 	return write(fd, pkt, sizeof(pkt)) == sizeof(pkt);
 }
 
-bool cairnsmore_supports_dynclock(int fd)
+bool cairnsmore_supports_dynclock(int fd, int read_size)
```
luke-jr: Cairnsmore never has non-4 read_size

driver-cairnsmore.c
```diff
@@ -92,7 +95,8 @@ bool cairnsmore_supports_dynclock(int fd)
 			.work_restart = false,
 			.work_restart_notifier = {-1, -1},
 		};
-		icarus_gets((unsigned char*)&nonce, fd, &tv_finish, &dummy, 1);
+		icarus_gets(nonce_bin, fd, &tv_finish, &dummy, 1, read_size);
+		memcpy((char *)&nonce, nonce_bin, sizeof(nonce));
```
luke-jr: Cast to char\* is unnecessary.

driver-erupter.c
```diff
@@ -33,6 +33,7 @@ static bool _erupter_detect_one(const char *devpath, struct device_drv *drv)
 		.Hs = ERUPTER_HASH_TIME,
 		.timing_mode = MODE_DEFAULT,
 		.continue_search = true,
+		.read_size = 4,
```
luke-jr: Rather than setting this in every driver, it should just be a default in driver-icarus.

driver-antminer.c
nwoolls: It should be a standard CRC-5 algorithm, but it was taken from the BITMAINtech driver (credit in the header). I didn't see another CRC-5 implementation yet in the bfgminer codebase.

1am3r:
I'll setup debian on my laptop and test it there. SSH access is no problem. I'm running the debian arm hard float (armhf) machine, I'm wondering if it has to do something with that...

driver-icarus.c
```diff
@@ -557,6 +553,32 @@ static void get_options(int this_option_offset, struct ICARUS_INFO *info)
...
+	excess_size++;
+	
+	unsigned char excess_bin[excess_size];
+	memset(excess_bin, 0, sizeof(excess_bin));
+	// Read excess_size from Icarus
+	struct timeval tv_now;
+	icarus_gets(excess_bin, fd, &tv_now, NULL, 1, excess_size);
+	// Number of bytes that were still available
+	
+	// Is there a cooler way to do this? strlen() didn't work
```
luke-jr: Maybe icarus_gets should return actual bytes read?

driver-icarus.c
```diff
@@ -734,9 +776,11 @@ static bool icarus_init(struct thr_info *thr)
 			"BFG\0\x64\x61\x01\x1a\xc9\x06\xa9\x51\xfb\x9b\x3c\x73";
 		
 		icarus_write(fd, pkt, sizeof(pkt));
-		if (ICA_GETS_OK == icarus_gets((unsigned char*)&res, fd, &tv_finish, NULL, info->read_count))
+		memset(res_bin, 0, sizeof(res_bin));
+		if (ICA_GETS_OK == icarus_gets(res_bin, fd, &tv_finish, NULL, info->read_count, info->read_size)) {
```
luke-jr: Please keep braces on their own line.

driver-icarus.c
```diff
@@ -1098,8 +1154,9 @@ static int64_t icarus_scanhash(struct thr_info *thr, struct work *work,
 		       icarus->proc_repr,
 		       (uint64_t)estimate_hashes,
 		       (int64_t)elapsed.tv_sec, (unsigned long)elapsed.tv_usec);
-
+
```
luke-jr: Please don't change formatting on its own.

driver-antminer.c
nwoolls: As far as I could tell this is not a straight hex conversion but is a lookup table of some sort. I didn't see a way to allow the user to enter the desired clock and send that to the device.

You can see the table of hex values and corresponding clocks here: https://github.com/AntMiner/AntGen1/blob/master/cgminer/README.md

driver-antminer.c
luke-jr: There are 3 different CRC-5 standards: https://en.wikipedia.org/wiki/Cyclic_redundancy_check#Commonly_used_and_standardized_CRCs

Ideal if you can identify which one so the function can later be moved to util.c if shared.

driver-cairnsmore.c
nwoolls: 10-4, I'll remove that and the other 3 uses.

driver-erupter.c
nwoolls: It's already set in Icarus, I'll remove it from the others. I was just trying to be explicit / readable.

driver-cairnsmore.c
nwoolls: True - but what would you suggest? I don't want to hard-code a 4 there when it is already specified in ICARUS_INFO->read_size. Seems bad practice to have two areas that say "4" for a driver when 1 would do.

driver-antminer.c
luke-jr: Looks like 5 or 6 if clauses could handle these..

driver-cairnsmore.c
luke-jr: #define CAIRNSMORE_READ_SIZE perhaps

driver-antminer.c
nwoolls: This is CRC5-USB. Do you want me to rename it and move it to util.c?

driver-antminer.c
nwoolls: True we could do our own lookup table, but then users would have to refer to documentation anyway to see what valid clock rates are. This way plays nicely with the BITMAINtech documentation.

What if there are more hex codes that just aren't documented? And then, when they are revealed, a new build of bfgminer has to be compiled to add them to bfgminer's lookup table to match BITMAINtech's lookup table.

I agree that I'd rather be able to enter a clock rate, I'm just not sure of the cleanest way to do it.

driver-cairnsmore.c
nwoolls: But the 4 isn't in the cairnsmore driver, its in the Icarus driver :)

driver-antminer.c
luke-jr: That sounds best. Please try to match general location (I think the very bottom) with driver-hashfast.c in the hashfast branch.

driver-antminer.c
luke-jr: If the setting begins with "x" parse it this way perhaps? Or maybe someone at bitmain can explain the values better?

driver-icarus.c
nwoolls: That would probably be preferable but I think would take someone more versed in the codebase than myself. I don't have the epoll requirements to test that codepath and am just not familiar enough with the inner workings of Icarus to start changing those things.

I do agree that would be cleaner though.

driver-cairnsmore.c
luke-jr: Hmm, just put a 4 here then. Not worth spending too much time on..

luke-jr: Be sure to check it == 4 in the driver-icarus before calling the func too, of course.

nwoolls: Taking a step back: one of the larger changes for this driver was to remove all of the hard coded defines for an Icarus-based driver's read-size and to instead specify it on the Icarus info struct. That is the basis of making the AntMiner U1 work.

What you are suggesting is essentially a step backwords: to go back to using hard-coded defines to specify a driver's read-size. In addition, the value for Cairnsmore (4) is already in the code once. So what you are suggesting is defining the read-size of Cairnsmore twice in code in two separate locations, which could lead to bugs in the future.

jimihendrix857:
Thanks for the quick response. So the antminer u1 devices should be setup with the winusb driver or the sillabs driver to work properly? Cause they are currently setup as sillabs and as soon as I type in the -S all the program crashes

Any help will be appreciated

Thanks in advance

driver-icarus.c
```diff
@@ -557,6 +553,33 @@ static void get_options(int this_option_offset, struct ICARUS_INFO *info)
...
+	// Try to read one more to ensure the device doesn't return
+	// more than we want for this driver
+	excess_size++;
+	
+	unsigned char excess_bin[excess_size];
+	memset(excess_bin, 0, sizeof(excess_bin));
+	// Read excess_size from Icarus
+	struct timeval tv_now;
+	cgtime(&tv_now);
+	icarus_gets(excess_bin, fd, &tv_now, NULL, 1, excess_size);
```
luke-jr: Why use icarus_gets at all? Just read is probably fine?

driver-antminer.c
```diff
@@ -0,0 +1,178 @@
+/*
+ * Copyright 2013 Luke Dashjr
+ * Copyright 2013 Nate Woolls
+ * Copyright 2013 Lingchao Xu <lingchao.xu@bitmaintech.com>
```
luke-jr: Please put the email address in AUTHORS only.

driver-antminer.c
```diff
@@ -0,0 +1,178 @@
...
+#include "lowl-vcom.h"
+
+#define ANTMINER_IO_SPEED 115200
+#define ANTMINER_HASH_TIME 0.0000000004761
+
+#define ANTMINER_STATUS_LEN 5
+
+BFG_REGISTER_DRIVER(antminer_drv)
+
+static bool antminer_detect_one(const char *devpath)
```
luke-jr: Keywords like static, inline, etc should be on a separate line please.

driver-antminer.c
```diff
@@ -0,0 +1,178 @@
...
+	
+	struct timeval tv_now;
+	
+	rdreg_buf[0] = 4;
+	rdreg_buf[0] |= 0x80;
+	rdreg_buf[1] = 0;    //16-23
+	rdreg_buf[2] = 0x04; // 8-15
+	rdreg_buf[3] = crc5usb(rdreg_buf, 27);
+	
+	applog(LOG_DEBUG, "Get clock: %02x%02x%02x%02x", rdreg_buf[0], rdreg_buf[1], rdreg_buf[2], rdreg_buf[3]);
```
luke-jr: Please prepend with "%"PRIpreprv": ", proc_repr

driver-antminer.c
```diff
@@ -0,0 +1,178 @@
...
+	
+	cgtime(&tv_now);
```
luke-jr: timer_set_now is the BFGMiner norm for clarity

driver-antminer.c
```diff
@@ -0,0 +1,178 @@
...
+	int err = icarus_write(cgpu->device_fd, rdreg_buf, sizeof(rdreg_buf));
+	
+	if (err != 0) {
```
luke-jr: \n {

driver-antminer.c
```diff
@@ -0,0 +1,178 @@
...
+	return NULL;
+}
+
+static
+char *antminer_set_clock(struct cgpu_info *cgpu, char *setting, char *replybuf)
+{
+	// For now we only allow hex values that use BITMAINtech's lookup table
+	// This means values should be prefixed with an x so that later we can
+	// accept and distinguish decimal values
+	if (strlen(setting) <= 1 || setting[0] != 'x') {
```
luke-jr: strlen is redundant and unnecessary.

driver-antminer.c
```diff
@@ -0,0 +1,178 @@
...
+		sprintf(replybuf, "invalid clock: '%s' data must be prefixed with an x", setting);
+		return replybuf;
+	}
+	
+	//remove leading character
+	char *hex_setting = setting + 1;
+	
+	if (strlen(hex_setting) > 8 || strlen(hex_setting) % 2 != 0 || strlen(hex_setting) / 2 == 0) {
```
luke-jr: hex2bin should check this for you

driver-antminer.c
```diff
@@ -0,0 +1,178 @@
...
+	antminer_drv.dname = "antminer";
+	antminer_drv.name = "AMU";
+	antminer_drv.lowl_probe = antminer_lowl_probe;
+	antminer_drv.set_device = antminer_set_device,
+	++antminer_drv.probe_priority;
+}
+
+struct device_drv antminer_drv = {
+	.drv_init = antminer_drv_init,
+};
```
luke-jr: Missing newline at EOF

driver-cairnsmore.c
luke-jr: #define CAIRNSMORE_READ_SIZE 4 can be in icarus-common.h since it is needed in both places.

driver-icarus.c
```diff
@@ -734,9 +777,12 @@ static bool icarus_init(struct thr_info *thr)
...
 		
 		icarus_write(fd, pkt, sizeof(pkt));
-		if (ICA_GETS_OK == icarus_gets((unsigned char*)&res, fd, &tv_finish, NULL, info->read_count))
+		memset(res_bin, 0, sizeof(res_bin));
+		if (ICA_GETS_OK == icarus_gets(res_bin, fd, &tv_finish, NULL, info->read_count, info->read_size))
+		{
+			memcpy(&res, res_bin, sizeof(res));
 			res = be32toh(res);
-		else
+		} else
```
luke-jr: \n

driver-antminer.c
nwoolls: It's safe to access setting[0] without checking the length first? What makes it redundant - the check in antminer_set_device? If so I'll move that to antminer_set_clock for clarity.

driver-cairnsmore.c
nwoolls: So you want me to initialize all Icarus drivers with CAIRNSMORE_READ_SIZE ?

nwoolls: (hint this is what you are saying)

nwoolls: Also please tell me why you are fighting tooth and nail for this? Again why would you want to specify the read-size of an Icarus based driver in 2 places instead of 1? You are still suggesting that. It would be defined in the DEFINE and in the ICARUS_INFO->read_size.

driver-antminer.c
luke-jr: setting[0] always exists in a null-terminated string. If the length is 0, setting[0] is '\0'.

driver-cairnsmore.c
luke-jr: There's just no point degrading the performance of code where there is no variability. ICARUS_INFO->read_size can initialise like usual (default 4), but driver-icarus.c should be checking read_size==4 before it calls this function.

driver-antminer.c
```diff
@@ -0,0 +1,178 @@
...
+ */
+
+#include "config.h"
+
+#include <stdbool.h>
+
+#include "miner.h"
+#include "icarus-common.h"
+#include "lowlevel.h"
+#include "lowl-vcom.h"
```
luke-jr, line 24: Missing includes:
stdint.h
stdio.h
stdlib.h
string.h
strings.h

"deviceapi.h"
"logging.h"
"util.h"

nwoolls, line 24: How did you work this part out? None of the IDEs or compilers I used showed any warnings or errors for this. Is there something I can run that warns on these?

driver-cairnsmore.c
nwoolls: do_default_detection won't be true, so there is no performance penalty.

nwoolls: Or delete the Cairnsmore driver since the thing doesn't exist and toss this whole discussion out?

driver-antminer.c
luke-jr, line 24: I read through the file and looked up the definition of each function/type used.

driver-cairnsmore.c
luke-jr: huh?

driver-icarus.c
```diff
@@ -1110,7 +1178,7 @@ static int64_t icarus_scanhash(struct thr_info *thr, struct work *work,
 	else
 		inc_hw_errors(thr, state->last_work, nonce);
 	icarus_transition_work(state, work);
-
+
```
luke-jr: .

driver-icarus.c
```diff
@@ -933,8 +994,7 @@ static int64_t icarus_scanhash(struct thr_info *thr, struct work *work,
 	int ret;
 
 	struct ICARUS_INFO *info;
-
-	uint32_t nonce;
+
```
luke-jr: .

driver-icarus.c
```diff
@@ -616,6 +653,15 @@ bool icarus_detect_custom(const char *devpath, struct device_drv *api, struct IC
 			devpath, nonce_hex, golden_nonce);
 		return false;
 	}
+		
+	if (info->read_size - ICARUS_NONCE_SIZE != bytes_left) {
```
luke-jr: \n

driver-antminer.c
```diff
@@ -0,0 +1,189 @@
...
+BFG_REGISTER_DRIVER(antminer_drv)
+
+static
+bool antminer_detect_one(const char *devpath)
+{
+	struct device_drv *drv = &antminer_drv;
+	
+	struct ICARUS_INFO *info = calloc(1, sizeof(struct ICARUS_INFO));
+	if (unlikely(!info))
+	quit(1, "Failed to malloc ICARUS_INFO");
```
luke-jr: indent

driver-antminer.c
```diff
@@ -0,0 +1,189 @@
...
+		return replybuf;
+	}
+	
+	applog(LOG_DEBUG, "%"PRIpreprv": Get clock: OK", cgpu->proc_repr);
+	
+	memset(rebuf, 0, sizeof(rebuf));
+	err = icarus_gets(rebuf, cgpu->device_fd, &tv_now, NULL, 10, ANTMINER_STATUS_LEN);
+	
+	// Timeout is ok - checking specifically for an error here
+	if (err == -1)
```
luke-jr: ICA_GETS_ERROR

driver-cairnsmore.c
```diff
@@ -84,7 +84,10 @@ bool cairnsmore_supports_dynclock(int fd)
 		return false;
 	if (!cairnsmore_send_cmd(fd, 0, 1, true))
 		return false;
-
+
```
luke-jr: .

driver-icarus.c
nwoolls: What? Nothing to fix here?

driver-antminer.c
```diff
@@ -0,0 +1,189 @@
...
+bool antminer_lowl_probe(const struct lowlevel_device_info * const info)
+{
+	return vcom_lowl_probe_wrapper(info, antminer_detect_one);
+}
+
+static
+char *antminer_get_clock(struct cgpu_info *cgpu, char *replybuf)
+{
+	uint8_t rdreg_buf[4] = {0};
+	uint8_t reg_data[4] = {0};
```
luke-jr: This isn't used...

nwoolls: What isn't used?

driver-icarus.c
```diff
@@ -557,6 +555,34 @@ static void get_options(int this_option_offset, struct ICARUS_INFO *info)
...
+	unsigned char excess_bin[excess_size];
+	memset(excess_bin, 0, sizeof(excess_bin));
+	// Read excess_size from Icarus
+	struct timeval tv_now;
+	timer_set_now(&tv_now);
+	//icarus_gets(excess_bin, fd, &tv_now, NULL, 1, excess_size);
+	read(fd, excess_bin, excess_size);
+	// Number of bytes that were still available
+	
+	// Is there a cooler way to do this? strlen() didn't work
```
luke-jr: Return value of read should be used here.

luke-jr: Note memsetting excess_bin is also unnecessary once this change is made.

driver-icarus.c
luke-jr: The blank line changed whitespace.

driver-antminer.c
luke-jr: The reg_data array.

driver-cairnsmore.c
```diff
@@ -85,6 +85,9 @@ bool cairnsmore_supports_dynclock(int fd)
 	if (!cairnsmore_send_cmd(fd, 0, 1, true))
 		return false;
 
+	// For reading the nonce from Icarus
+	unsigned char nonce_bin[ICARUS_DEFAULT_READ_SIZE];
+	// For storing the 32-bit nonce
```
luke-jr: No need for nonce_bin refactor here.

PR #376, opened 2014-01-13
Add Difficulty adjustment countdown in terms of

TrooperT:
blocks remaining until next adjustment

### DIFF
--- a/README
+++ b/README
@@ -431,11 +431,12 @@ dedicated to this program,
 	https://bitcointalk.org/?topic=168174
 
 The block display shows:
-Block: ...1b89f8d3 #217364  Diff:7.67M (54.93Th/s)  Started: [17:17:22]
+Block: ...1b89f8d3 #217364 Diff:7.67M(54.93Th/s) TG:364 Started:[17:17:22]
 
 This shows a short stretch of the current block, the next block's height and
-difficulty (including the network hashrate that difficulty represents), and when
-the search for the new block started.
+difficulty (including the network hashrate that difficulty represents), block 
+remaining until next difficulty adjustment, and when the search for the new 
+block started.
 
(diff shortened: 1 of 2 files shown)


luke-jr:
This probably overflows the line for most terminals. Can we fit it nicer somehow? Also needs README updated.

TrooperT:
will think of a two letter and update. 

but do you like the idea?

luke-jr:
Sure. I wonder if it belongs after the block diff?

TrooperT:
i agree. will move around. sorry in advance for so many commit changes. working from a mobile phone

TrooperT:
there :)

luke-jr:
How does this react if the block height is not known (maybe test with an old old altcoin)? ;)

TrooperT:
was it a trick question? tested with no pool and --benchmark and no peculiar behavior

luke-jr:
I'm not aware of an easy way to test this, short of setting up a test pool... meet me on IRC if you want me to run one for you.

Right now, this looks like:

```
Block: ...89a64718e19ffe25 Diff:1( 7.16Mh/s) TG:2016 Started:[18:11:53]
```

TG should probably just not be shown at all in this case, though.

Observe the current behaviour:

```
Unknown block height: Block: ...d24b1f3c64df2faf  Diff:1.79G (12.81Ph/s)  Started: [18:10:58]
Known   block height: Block: ...214dbffd #280331  Diff:1.79G (12.81Ph/s)  Started: [18:04:36]
```

luke-jr:
@Taugeran ping?

TrooperT:
closing for now. just had a nightmare spring to life IRL. barely keeping self together

luke-jr:
:( hope things get better soon.

PR #388, opened 2014-02-03
Minor improvements

jesterKing:
A few minor improvements, the most important being the datestamp string publication in the API summary response.

### DIFF
--- a/api.c
+++ b/api.c
@@ -1280,6 +1280,7 @@ static void message(struct io_data * const io_data, const int messageid2, const 
 			}
 
 			root = api_add_string(root, _STATUS, severity, false);
+			root = api_add_escape(root, "Since", datestamp, false);
 			root = api_add_time(root, "When", &when, false);
 			root = api_add_int(root, "Code", &messageid, false);
 			root = api_add_escape(root, "Msg", buf, false);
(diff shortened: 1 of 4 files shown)


driver-antminer.c
```diff
@@ -167,13 +167,13 @@ char *antminer_set_clock(struct cgpu_info *cgpu, char *setting, char *replybuf)
 
 static
 char *antminer_set_device(struct cgpu_info *cgpu, char *option, char *setting, char *replybuf)
-{		
-	if (strcasecmp(option, "clock") == 0)
+{
+	if (strncasecmp(option, "clock", 5) == 0)
```
luke-jr, line 171: Why?

api.c
```diff
@@ -1280,6 +1280,7 @@ static void message(struct io_data * const io_data, const int messageid2, const
 			}
 
 			root = api_add_string(root, _STATUS, severity, false);
+			root = api_add_escape(root, "Since", datestamp, false);
```
luke-jr, line 1283: datestamp is the time when BFGMiner was started (or stats reset). I'm not sure it makes sense to include this in _every_ response. Additionally, since this is a programmic interface, it should be returned as a UNIX time, not a string.

luke-jr:
No explanations, closing.

PR #396, opened 2014-02-10
opencl: Allow specifying a comma separated list for --gpu-threads

nwoolls:


driver-opencl.c
```diff
@@ -486,6 +486,21 @@ const char *set_intensity(char *arg)
...
+	if (v < 1 || v > 10)
+		return false;
+	struct opencl_device_data * const data = cgpu->device_data;
+	cgpu->threads = v;
+	return true;
+}
+const char *set_gpu_threads(char *arg)
+{
+	return _set_list(arg, "Invalid value passed to set_gpu_threads", _set_gpu_threads);
+}
```
luke-jr: I think _SET_INT_LIST should fit for this?

miner.c
```diff
@@ -2063,10 +2063,10 @@ static char *set_null(const char __maybe_unused *arg)
 		     set_int_0_to_9999, opt_show_intval, &opt_platform_id,
 		     "Select OpenCL platform ID to use for GPU mining"),
 	OPT_WITH_ARG("--gpu-threads",
-		     set_int_1_to_10, opt_show_intval, &opt_g_threads,
-		     "Number of threads per GPU (1 - 10)"),
+		     set_gpu_threads, NULL, NULL,
+		     "Number of threads per GPU (1 - 10) - one value for all or separate by commas for per card"),
 	OPT_WITH_ARG("-g",
-	             set_int_1_to_10, opt_show_intval, &opt_g_threads,
+	             set_gpu_threads, NULL, NULL,
```
luke-jr: This removes the default from --help; intended?

README
```diff
@@ -299,7 +299,7 @@ GPU only options:
 
 --auto-fan          Automatically adjust all GPU fan speeds to maintain a target temperature
 --auto-gpu          Automatically adjust all GPU engine clock speeds to maintain a target temperature
---gpu-threads <arg> Number of threads per GPU (1 - 10) (default: -1)
+--gpu-threads <arg> Number of threads per GPU (1 - 10) (default: -1) - one value for all or separate by commas for per card.
```
luke-jr: Be sure this matches what is in --help

driver-opencl.c
nwoolls: I don't think so - almost. And very clever macro.

_SET_INT_LIST sets a field on the OpenCL device data structure. This is setting a field on the cgpu_info structure.

miner.c
nwoolls: Not sure what you mean here, the default of -1 is set in miner.c and opt_g_threads is still used if this argument isn't supplied.

driver-opencl.c
luke-jr: Would it perhaps be easier to modify the macro arg to decide whether to set cgpu->\* or data->*?

miner.c
luke-jr: I mean opt_show_intval appends "(default: -1)" to the --help line. Thought perhaps that is not useful anyway.

driver-opencl.c
nwoolls: Your call. Depends on how common this is. I don't want to complicate the macro but I'm fine with either way. Let me know.

miner.c
nwoolls: Got ya - clearer after going through the source here:

http://ccodearchive.net/info/opt.html

It was unintentional and I'm fixing it up.

luke-jr:
Merged

PR #402, opened 2014-02-16
bitfury: Use dynclock to adjust osc6_bits

ronmat77:


luke-jr:
The dynclock module assumes a linear multiplier. osc6_bits is, unfortunately, not that. :/

ronmat77:
HI, hopefully this makes it to luke-jr...

Icefurys , can't figure out how to get the theoretical 2.5 to 2.9.. but
there is no info on manufacturer except tge 'pencil' trick.  But the
interrwebs say bfgminer can do it.

Any thoughts?

Thanks in advance.

PR #409, opened 2014-02-25
dualminer: Support for mining both Scrypt and SHA + Scrypt using DualMiner

nwoolls:


dtbartle:
What's the 0.9V/1.2V hub fans thing? From looking at the datasheets, I know it's possible to select four different core voltages, and in theory, two of those can be selected with software.

configure.ac
```diff
@@ -414,7 +414,7 @@ if test "x$bitforce" = xyes; then
 fi
 AM_CONDITIONAL([HAS_BITFORCE], [test x$bitforce = xyes])
 
-driverlist="$driverlist icarus cairnsmore/icarus erupter/icarus antminer/icarus"
+driverlist="$driverlist icarus cairnsmore/icarus erupter/icarus antminer/icarus dualminer/icarus"
```
luke-jr: This adds enough code that I'd like to see it available separately from icarus. See how the bitfury-based drivers depend on the main bitfury code.

driver-dualminer.c
```diff
@@ -0,0 +1,237 @@
...
+	if (opt_scrypt)
+		gc3355_set_rts_status(fd, RTS_HIGH);
+
+	gc3355_init(fd, opt_dualminer_pll, opt_dualminer_btc_gating, opt_scrypt);
+
+	if (gc3355_get_cts_status(fd) != 1)
+	{
+		//LTC + SHA mode
+		if (opt_scrypt)
+			info->Hs = DUALMINER_SCRYPT_HASH_TIME * 2;
```
luke-jr: Is it exactly 2 times slower? Maybe another constant for this is appropriate.

driver-dualminer.c
```diff
@@ -0,0 +1,237 @@
...
+#endif
+
+#define DUALMINER_IO_SPEED 115200
+
+#define DUALMINER_SCRYPT_HASH_TIME 0.00001350
+
+#define DUALMINER_SHA2_HASH_TIME 0.000000001950
+
+#define RTS_LOW 0
+#define RTS_HIGH 1
```
luke-jr: This should probably be an enum.

driver-dualminer.c
```diff
@@ -0,0 +1,237 @@
...
+#define LTC_UNIT_CLOSE 1
+
+BFG_REGISTER_DRIVER(dualminer_drv)
+
+static
+bool dualminer_detect_init(const char *devpath, int fd)
+{
+	gc3355_dual_reset(fd);
+
+	opt_scrypt ? gc3355_opt_ltc_only_init(fd) : gc3355_dualminer_init(fd);
```
luke-jr: Would prefer to see scrypt_only as an option to gc3355_dualminer_init

driver-dualminer.c
```diff
@@ -0,0 +1,237 @@
...
+	struct ICARUS_INFO *info = calloc(1, sizeof(struct ICARUS_INFO));
+	if (unlikely(!info))
+		quit(1, "Failed to malloc ICARUS_INFO");
+
+	*info = (struct ICARUS_INFO){
+		.baud = DUALMINER_IO_SPEED,
+		.timing_mode = MODE_DEFAULT,
+		.reopen_mode = IRM_NEVER,
+		.do_icarus_timing = false,
+		.reverse_nonce = true,
```
luke-jr: .nonce_bigendian preferred for clarity

driver-dualminer.c
```diff
@@ -0,0 +1,237 @@
...
+		"711c0000603ebdb6e35b05223c54f815"
+		"5ac33123006b4192e7aafafbeb9ef654"
+		"4d2973d700000002069b9f9e3ce8a677"
+		"8dea3d7a00926cd6eaa9585502c9b83a"
+		"5601f198d7fbf09be9559d6335ebad36"
+		"3e4f147a8d9934006963030b4e54c408"
+		"c837ebc2eeac129852a55fee1b1d88f6"
+		"000c050000000600";
+
+		info->golden_nonce = "00050cdd";
```
luke-jr, line 132: Can we replace these hex blobs with uint8_t arrays and/or bytes_t?

driver-dualminer.c
```diff
@@ -0,0 +1,237 @@
...
+	if (!icarus_detect_custom(devpath, drv, info))
+	{
+		free(info);
+		return false;
+	}
+
+	if (opt_scrypt)
+		info->read_count = 48; //4.8s to read
+	else
+		info->read_count = 16; //1.6s to read
```
luke-jr, line 158: These should be set before detect, or the user cannot override.

driver-dualminer.c
```diff
@@ -0,0 +1,237 @@
...
+
+static
+bool dualminer_job_prepare(struct thr_info *thr, struct work *work, __maybe_unused uint64_t max_nonce)
+{
+	struct cgpu_info * const icarus = thr->cgpu;
+	struct icarus_state * const state = thr->cgpu_data;
+	struct ICARUS_INFO * const info = icarus->device_data;
+	int fd = icarus->device_fd;
+
+	memset(state->ob_bin, 0, info->work_size);
```
luke-jr, line 191: This is going to be reassigned immediately, right?

driver-dualminer.c
```diff
@@ -0,0 +1,237 @@
...
+		uint8_t temp_bin[64];
+		memset(temp_bin, 0, 64);
+		memcpy(temp_bin, work->midstate, 32);
+		memcpy(temp_bin+52, work->data + 64, 12);
+		state->ob_bin[0] = 0x55;
+		state->ob_bin[1] = 0xaa;
+		state->ob_bin[2] = 0x0f;
+		state->ob_bin[3] = 0x00;
+		memcpy(state->ob_bin + 8, temp_bin, 32);
+		memcpy(state->ob_bin + 40, temp_bin + 52, 12);
```
luke-jr, line 218: What is the point of an extra copy to temp_bin first?

util.c
```diff
@@ -3052,3 +3050,88 @@ uint8_t crc8ccitt(const void * const buf, const size_t buflen)
 		crc = _crc8ccitt_table[crc ^ *p++];
 	return crc;
 }
+
+static
+unsigned char hex_str[2048];
+
+void print_hex(unsigned char *data, int len, const unsigned char * prefix)
```
luke-jr: IMO this should either be made generic and used in all drivers doing it or removed.

driver-icarus.c
```diff
@@ -231,6 +231,7 @@ int icarus_gets(unsigned char *buf, int fd, struct timeval *tv_finish, struct th
 		{
 			if (epollfd != -1)
 				close(epollfd);
+			print_hex((char *)buf, read_size, "Read from UART:\n");
```
luke-jr: This should only print with opt_dev_protocol_dump, and not using newlines...

driver-icarus.c
```diff
@@ -493,7 +474,40 @@ bool icarus_detect_custom(const char *devpath, struct device_drv *api, struct IC
...
+		 */
+
+		info->golden_nonce = "000187a2";
+		info->work_size = 64;
+	}
+
+	if (info->detect_init_func != NULL)
+	{
+		info->detect_init_func(devpath, fd);
+		usleep(1000);
```
luke-jr: sleep can be part of the custom detect_init_func.

driver-icarus.c
```diff
@@ -542,7 +563,11 @@ bool icarus_detect_custom(const char *devpath, struct device_drv *api, struct IC
 	icarus = calloc(1, sizeof(struct cgpu_info));
 	icarus->drv = api;
 	icarus->device_path = strdup(devpath);
-	icarus->device_fd = -1;
+	if (info->reopen_mode == IRM_NEVER)
```
luke-jr: IRM_NEVER is already used to mean something else. Please add a new option.

driver-icarus.c
```diff
@@ -755,8 +796,8 @@ static bool icarus_job_start(struct thr_info *thr)
 	}
 
 	if (opt_debug) {
-		char ob_hex[129];
-		bin2hex(ob_hex, ob_bin, 64);
+		char ob_hex[(sizeof(ob_bin) * 2) + 1];
```
luke-jr: ob_bin is a pointer, sizeof won't work right here.

driver-icarus.c
```diff
@@ -887,7 +928,10 @@ static int64_t icarus_scanhash(struct thr_info *thr, struct work *work,
 	struct icarus_state *state = thr->cgpu_data;
 	was_first_run = state->firstrun;
 
-	icarus_job_prepare(thr, work, max_nonce);
+	if (icarus->drv->job_prepare != NULL)
+		icarus->drv->job_prepare(thr, work, max_nonce);
+	else
+		icarus_job_prepare(thr, work, max_nonce);
```
luke-jr: Just set icarus_drv.job_prepare = icarus_job_prepare instead of checking this every scanhash

driver-icarus.c
```diff
@@ -934,6 +978,9 @@ static int64_t icarus_scanhash(struct thr_info *thr, struct work *work,
 				case ICA_GETS_OK:
 					break;
 			}
+
+			if (info->reverse_nonce)
+				rev(nonce_bin, 4);
```
luke-jr: Instead, change be32toh to le32toh if nonce_bigendian is false.

gc3355.c
```diff
@@ -0,0 +1,905 @@
...
+char *opt_dualminer_pll = NULL;
+char *opt_dualminer_btc_gating = NULL;
+int opt_pll_freq = 400;
+int opt_btc_number = 160;
+bool opt_hubfans = false;
+
+void gc3355_dual_reset(int fd)
+{
+	static int i=0;
+	applog(LOG_DEBUG,"--->>>%s():%d\n",__FUNCTION__, i++);
```
luke-jr: Don't use static variables like this. Probably safe to remove anyway?

gc3355.c
```diff
@@ -0,0 +1,905 @@
...
+#else
+
+	int dtr_flag = 0;
+	ioctl(fd, TIOCMGET, &dtr_flag);
+	dtr_flag |= TIOCM_DTR;
+	ioctl(fd, TIOCMSET, &dtr_flag);
+	usleep(1000);
+	ioctl(fd, TIOCMGET, &dtr_flag);
+	dtr_flag &= ~TIOCM_DTR;
+	ioctl(fd, TIOCMSET, &dtr_flag);
```
luke-jr, line 365: This all could be much simpler with a lowl-vcom DTR function.

gc3355.c
```diff
@@ -0,0 +1,905 @@
...
+		applog(LOG_DEBUG, "GC3355: set freq %s, reg2=%s in the support list\n", pll_freq, pll_freq_cmd[found_pll + 2]);
+		hex2bin(pllob_bin, pll_freq_cmd[found_pll + 2], sizeof(pllob_bin));
+		icarus_write(fd, pllob_bin, 8);
+		usleep(1000);
+	}
+	else
+		applog(LOG_ERR, "GC3355: freq %s is not supported\n", pll_freq);
+}
+
+int gc3355_get_cts_status(int fd)
```
luke-jr, line 492: Please make this lowl-vcom generic.

gc3355.c
```diff
@@ -0,0 +1,905 @@
...
+	return ret;
+#else
+	ioctl(fd, TIOCMGET, &status);
+	ret = (status & 0x20) ? 0 : 1;
+	applog(LOG_DEBUG, "Get CTS Status is : %d [Linux: 1 is 1.2; 0 is 0.9]\n", ret);
+	return ret;
+#endif
+}
+
+void gc3355_set_rts_status(int fd, unsigned int value)
```
luke-jr, line 509: lowl-vcom again

miner.c
```diff
@@ -1781,7 +1781,7 @@ static char *set_icarus_options(const char *arg)
 	char *opts = strdup(arg), *argdup;
 	argdup = opts;
 	const struct bfg_set_device_definition *sdf = icarus_set_device_funcs;
-	const char *drivers[] = {"antminer", "cairnsmore", "erupter", "icarus"};
+	const char *drivers[] = {"dualminer", "antminer", "cairnsmore", "erupter", "icarus"};
```
luke-jr: No, I intentionally named drivers here that predated the deprecation of --icarus-options. We don't want new drivers to be affected by them.

miner.c
```diff
@@ -8623,7 +8623,7 @@ void set_target(unsigned char *dest_target, double diff)
 	
 	if (opt_debug) {
 		char htarget[65];
-		bin2hex(htarget, rtarget, 32);
+		bin2hex(htarget, dest_target, 32);
```
luke-jr: Why? This is wrong...

util.c
```diff
@@ -689,42 +689,40 @@ void bin2hex(char *out, const void *in, size_t len)
...
+		p++;
+		hexstr += 2;
+		len--;
 	}
-	
-	return likely(!hexstr[0]);
+
+	if (likely(len == 0 && *hexstr == 0))
+		ret = true;
+	return ret;
```
luke-jr, line 725: Why are you replacing BFGMiner's optimised hex2bin with cgminer's much slower version?? :/

util.c
```diff
@@ -3052,3 +3050,88 @@ uint8_t crc8ccitt(const void * const buf, const size_t buflen)
 		crc = _crc8ccitt_table[crc ^ *p++];
 	return crc;
 }
+
+static
+unsigned char hex_str[2048];
```
luke-jr: There is no reason to have this static, and it isn't thread-safe...

util.c
```diff
@@ -3052,3 +3050,88 @@ uint8_t crc8ccitt(const void * const buf, const size_t buflen)
...
+void print_hex(unsigned char *data, int len, const unsigned char * prefix)
+{
+    int i, j, s, blank;
+    unsigned char *p = data;
+    unsigned char *ptr = hex_str;
+
+    memset(hex_str, 0, sizeof(hex_str));
+    if(prefix == NULL)
+	{
+        sprintf(ptr, "\n", prefix);
```
luke-jr: Better to not use newlines in logging.

luke-jr:
Left a lot of suggestions. Not all of them _need_ to be fixed before merging, but some are notable bugs.

driver-dualminer.c
nwoolls, line 158: Not sure I agree here - or I am misunderstanding the code. Maybe you can step through with your DualMiner. If you look at icarus_set_timing() which is always called, the DEFAULT branch for that resets read_count.  Basically there is no code path (that I see) where I can set the info->read_count, then call icarus_detect_custom() and have the info->read_count come out the same on the other end.

The AntMiner driver does the same thing.

driver-dualminer.c
nwoolls, line 191: Yes but not the entire size of the state->ob_bin. That bin is now 160 elements to allow for the larger work. I am filling with 0s first as it may only be 52 elements written.

util.c
nwoolls: Ya I was also thinking maybe it should require a new level of --debug like --dev-debug. I think this is super useful (I used something similar with the AntMiner code) so I hate to keep removing it.

driver-icarus.c
nwoolls: And then you go and answer my question / comments from above ;) I'll check that option.

nwoolls: About the newline though - can you try this out with one of your Icarus devices or the DualMiner? It outputs the information in columns - not using a newline would throw off the first row.

Example:
 [2014-03-01 13:05:08] Send to UART:
0000 : 46 79 ba 4e c9 98 76 bf - 4b fe 08 60 82 b4 00 25    Fy.N..v.K..`...%
0010 : 4d f6 c3 56 45 14 71 13 - 9a 3a fa 71 e4 8f 54 4a    M..VE.q..:.q..TJ
0020 : 00 00 00 00 00 00 00 00 - 00 00 00 00 00 00 00 00    ................
0030 : 00 00 00 00 87 32 0b 1a - 14 26 67 4f 2f a7 22 ce    .....2...&gO/.".

util.c
nwoolls: See my example output above - the output is broken into rows / columns

util.c
nwoolls, line 725: Because it doesn't work with the hex codes in the GridSeed reference code. If you can debug with your DualMiner and figure out why leaving the old code would be preferred. This part was over my head.

gc3355.c
nwoolls, line 365: Do you have time to do these lowl-vcom changes? Unfortunately I do not have a device for testing. I've made the rest of the changes.

nwoolls:
Please take a look at this:

https://github.com/nwoolls/bfgminer/commits/temp/dualminer-rewrite-commits2

Let me know how that looks along with my comment replies above, and then I'll go from there.

driver-dualminer.c
luke-jr, line 191: It would probably be clearer to memset only the regions which will be zero.

driver-dualminer.c
luke-jr, line 158: If info->Hs is set correctly, this should come out to the right value...

PR #414, opened 2014-03-07
dualminer: Support for mining both Scrypt and SHA2 using DualMiner GC3355 thumb sticks

nwoolls:
Set clock frequency using --set-device dualminer:clock=frequency (default is 850 in Scrypt mode)
Mine Scrypt2 and SHA2 simultaneously with two miner processes using --dual-mode (SHA2 process must be launched first)


luke-jr:
I've refactored this to the dualminer branch on github. Haven't tested yet.

PR #416, opened 2014-03-10
Drillbit support for new protocol revision, miniplane connected boards

projectgus:
Hi Luke,

We have a new Drillbit firmware revision in beta release that supports miniplane support (joining multiple boards into a single unit), some details at http://drillbitsystem.com/forum/index.php?topic=307.0

It also brings a new configuration packet format, which is less ASIC-specific (designed to accommodate our forthcoming Avalon-based miners.) Where we previously had fields for "int clock level" and "ext clock freq", we just have a single frequency field now. For Bitfury "internal" clock levels, there's some new logic in the firmware. Frequencies less than 64 are assumed to be clock levels and set directly. Values >=64 are divided by 5 to convert frequency to level, approximately. I've left the Bfgminer settings alone though, the change is fully "behind the scenes" and settings are defined the same as before.

Commit 4acf605 has a bit more detail about that small tweak in the commit message.

Thanks very much for all your hard work with bfgminer. It's really slick. :)

Please let me know if there's anything you'd like me to change.

Angus

### DIFF
--- a/driver-drillbit.c
+++ b/driver-drillbit.c
@@ -95,7 +95,7 @@ err:
 			strcpy(product, "Eight");
 	}
 	else
-	if ((chips % 8 == 0) && !strcmp(product, "Eight"))
+	if ((chips >= 8) && (chips <= 64) && (chips % 8 == 0) && !strcmp(product, "Eight"))
 	{}  // Known device
 	else
 	if (chips == 1 && !strcmp(product, "Thumb"))
--- a/driver-drillbit.c
+++ b/driver-drillbit.c
@@ -21,7 +21,7 @@
 BFG_REGISTER_DRIVER(drillbit_drv)
 
 #define DRILLBIT_MIN_VERSION 2
-#define DRILLBIT_MAX_VERSION 3
+#define DRILLBIT_MAX_VERSION 4
 
 #define DRILLBIT_MAX_WORK_RESULTS 0x400
 #define DRILLBIT_MAX_RESULT_NONCES 0x10
@@ -31,22 +31,15 @@ enum drillbit_capability {
 	DBC_EXT_CLOCK = 2,
 };
 
-enum drillbit_voltagecfg {
-	DBV_650mV = 0,
-	DBV_750mV = 2,
-	DBV_850mV = 1,
-	DBV_950mV = 3,
-};
-
 struct drillbit_board {
-	enum drillbit_voltagecfg core_voltage_cfg;
-	unsigned clock_level;
+	unsigned core_voltage;
+	unsigned clock_freq;
 	bool clock_div2;
 	bool use_ext_clock;
-	unsigned ext_clock_freq;
 	bool need_reinit;
 	bool trigger_identify;
 	uint16_t caps;
+	uint8_t protover;
 };
 
 static
@@ -84,7 +77,7 @@ err:
 	char * const product = (void*)&buf[1];
 	buf[9] = '\0';  // Ensure it is null-terminated (clobbers serial, but we already parsed it)
 	unsigned chips = buf[0xd];
-	uint16_t caps = (uint16_t)buf[0xe] | ((uint16_t)buf[0xf] << 8);
+	intptr_t caps = buf[0xe] | ((intptr_t)buf[0xf] << 8);
 	if (!product[0])
 		applogr(false, LOG_DEBUG, "%s: %s: Null product name", __func__, devpath);
 	if (!serialno)
@@ -121,6 +114,8 @@ err:
 		// Production firmware Thumbs don't set any capability bits, so fill in the EXT_CLOCK one
 		caps |= DBC_EXT_CLOCK;
 	
+        caps |= ((intptr_t)protover) << 16; // Store protocol version in capability field, temporarily
+
 	char *serno = malloc(9);
 	snprintf(serno, 9, "%08lx", serialno);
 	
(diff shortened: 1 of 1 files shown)


driver-drillbit.c
```diff
@@ -84,7 +77,7 @@ bool drillbit_detect_one(const char * const devpath)
 	char * const product = (void*)&buf[1];
 	buf[9] = '\0';  // Ensure it is null-terminated (clobbers serial, but we already parsed it)
 	unsigned chips = buf[0xd];
-	uint16_t caps = (uint16_t)buf[0xe] | ((uint16_t)buf[0xf] << 8);
+	intptr_t caps = buf[0xe] | ((intptr_t)buf[0xf] << 8);
```
luke-jr: Why is this changed to intptr_t (which is of undefined size)?

driver-drillbit.c
```diff
@@ -102,7 +95,7 @@ bool drillbit_detect_one(const char * const devpath)
 			strcpy(product, "Eight");
 	}
 	else
-	if (chips == 8 && !strcmp(product, "Eight"))
+	if ((chips % 8 == 0) && !strcmp(product, "Eight"))
```
luke-jr: ?

driver-drillbit.c
```diff
@@ -110,17 +103,18 @@ bool drillbit_detect_one(const char * const devpath)
...
 	
-	if (protover < DRILLBIT_MIN_VERSION || (loglev == LOG_DEBUG && protover > DRILLBIT_MAX_VERSION))
-		applogr(false, loglev, "%s: %s: Unknown device protocol version %u.",
-		        __func__, devpath, protover);
-	if (protover > DRILLBIT_MAX_VERSION)
-		applogr(false, loglev, "%s: %s: Device firmware uses newer Drillbit protocol %u. We only support up to %u. Find a newer BFGMiner!",
+	if (protover < DRILLBIT_MIN_VERSION || (loglev == LOG_DEBUG && protover > DRILLBIT_MAX_VERSION)) {
+		applogr(false, LOG_ERR, "%s: %s: Device firmware uses newer Drillbit protocol %u. We only support up to %u. Find a newer BFGMiner!",
 		        __func__, devpath, protover, (unsigned)DRILLBIT_MAX_VERSION);
+		return false;
```
luke-jr: applogr includes the return..

driver-drillbit.c
```diff
@@ -206,8 +200,34 @@ bool drillbit_send_config(struct cgpu_info * const dev)
...
+	uint8_t buf[7] = {'C'};
+	if(board->protover < 4) {
+		if(board->core_voltage < 750)
+			buf[1] = 0; // 650mV
+		else if(board->core_voltage < 850)
+			buf[1] = 1; // 750mV
+		else if(board->core_voltage < 950)
+			buf[1] = 2; // 850mV
+		else
+			buf[1] = 3; // 950mV
```
luke-jr: Why replace the enum with this?

driver-drillbit.c
projectgus: Hi Luke,

I think I was thinking along the lines of using intptr_t as we're storing this value in a pointer later on (and intptr_t is defined as size of a pointer.) Figured to save a cast further down, then I forgot to remove the cast!

I've rewritten to be a bit more explicit about what's going on in bb144a11, how is that?

driver-drillbit.c
projectgus: Eight boards in a miniplane show up as 8,16,24,etc. ASICs depending on size. Realised I hadn't put any bounds limits though, fixed in 84ed01be.

driver-drillbit.c
projectgus: _Facepalm_, thanks. I could have sworn I saw a V4 board connecting to an unpatched bfgminer and trying to hash, but I can't reproduce it now and obviously I didn't test the "fix" very well when I did it. Reverted in 4fc40215.

driver-drillbit.c
projectgus: Took it out as it was only used in one place now (here.) The reason for moving to integer voltages was we have some new boards coming out with different power supply models, and didn't want to have to keep bleeding the firmware implementation details into the miner software.

However that's no real need to remove the enum entirely, and in doing so I introduced a subtle bug (750mV and 850mV settings are swapped.) Reintroduced in 83b166d.

projectgus:
Thanks for the comments. If you prefer a clean history I can rebase this lot into a new branch and submit as a new PR.

driver-drillbit.c
```diff
@@ -32,21 +32,21 @@ enum drillbit_capability {
...
-	DBV_650mV = 0,
-	DBV_750mV = 2,
-	DBV_850mV = 1,
-	DBV_950mV = 3,
-};
+       DBV_650mV = 0,
+       DBV_750mV = 2,
+       DBV_850mV = 1,
+       DBV_950mV = 3,
+}; // Flags used for pre-V4 configuration format only
```
luke-jr, line 39: I guess I missed that v4 doesn't use these flags anymore... I guess go back to how you had it (removing the enum).

driver-drillbit.c
```diff
@@ -84,7 +84,7 @@ bool drillbit_detect_one(const char * const devpath)
 	char * const product = (void*)&buf[1];
 	buf[9] = '\0';  // Ensure it is null-terminated (clobbers serial, but we already parsed it)
 	unsigned chips = buf[0xd];
-	uint16_t caps = (uint16_t)buf[0xe] | ((uint16_t)buf[0xf] << 8);
+	uint16_t caps = buf[0xe] | ((uint16_t)buf[0xf] << 8);
```
luke-jr, line 87: I prefer to keep this explicit for clarity.

luke-jr:
Sorry for the delay, I'm back... Merged in 137f6d1 except for the enum reintroduction.

PR #417, opened 2014-03-11
Restore LTC cpu mining - currently it generates hw errors only

lumag:
Somehow LTC cpu mining got broken - restore it. There is no need to endian-convert generated nonce.


luke-jr:
Fixed in c8b9cc4 in a more optimal way (iteration of nonce is done big endian for improved performance). Thanks for reporting!

PR #424, opened 2014-04-01
gridseed: Added support for Scrypt hashing with 5-chip GridSeed devices

nwoolls:
AFAIK SHA-2 support _should_ work for the GridSeed devices, but doesn't. Given resources and timing I've flagged the driver Scrypt-only for now. Adjusting the flag for SHA would allow anyone else to continue work on SHA support. Running with --device-protocol-dump generates protocol logs that are seemingly identical to working logs.


luke-jr:
Merged via 39461ee.

PR #428, opened 2014-04-12
Update driver-nanofury.c

nanofury:
Add NANOFURY_GP_PIN_PWR_EN0 (inverted PWR_EN) - support for power management (hardware) optimizations

### DIFF
--- a/driver-nanofury.c
+++ b/driver-nanofury.c
@@ -27,6 +27,7 @@
 #define NANOFURY_GP_PIN_LED 0
 #define NANOFURY_GP_PIN_SCK_OVR 5
 #define NANOFURY_GP_PIN_PWR_EN 6
+#define NANOFURY_GP_PIN_PWR_EN0 7
 
 #define NANOFURY_MAX_BYTES_PER_SPI_TRANSFER 60			// due to MCP2210 limitation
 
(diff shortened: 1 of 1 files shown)


driver-nanofury.c
```diff
@@ -165,6 +166,10 @@ bool nanofury_checkport(struct mcp2210_device * const mcp, const unsigned long b
 	if (!mcp2210_set_gpio_output(mcp, NANOFURY_GP_PIN_PWR_EN, MGV_HIGH))
 		goto fail;
 	
+	// PWR_EN0
```
luke-jr, line 169: What is PWR_EN0? :)

nanofury, line 169: PWR_EN0 is an inverted power-enable signal (e.g. "active" is "0"). 
This signal serves the same purpose but has the opposite state of PWR_EN for which the "active" state is "1".

In essence this will permit more options for selecting the power-controlling circuits (as some require a "1" for "on" state and others require a "0" for the same).

For PWR_EN0 the two states are:
Output "0" - power is on (bitfury chips are powered)
Output "1" or Input - power is off (butfury chips are not powered)

luke-jr, line 169: So this is a PWR_EN _or_ PWR_EN0 thing? ie, no timing issue to worry about coordinating it with PWR_EN?

nanofury, line 169: Correct - no timing issue.

And It is AND (not OR). Always switch both signals into ON state.

At any given time only one of them will be used (but that will be depending on the specific hardware)  and there is no disadvantage to switch both signals. 

From a timing prospective - there are no limitations and it is not important which one goes into "ON" state first and which is second.

I would recommend also adding a 0.1sec delay after power has been turned on (so that voltage can stabilize) before any commands to the bitfury chips are sent.

luke-jr:
Merged as a65cdff

PR #430, opened 2014-04-14
opencl: Accept arbitrary kernel filenames, provided they specify a kernel interface somewhere in the source file

luke-jr:
File must include the string "kernel-interface:" followed by one of the known interface names


nwoolls:
I love you man.

OhGodAPet:
Thanks, Luke, for implementing my suggestion. I don't know if I could have done it.

luke-jr:
Can someone please test this, perhaps with some of sgminer's kernel files? They should just work as-is...

nwoolls:
I'm out of town again this week but will be happy to test when I get back.

PR #433, opened 2014-04-20
Added scripts for easier startup for shells (Linux/Uni*).

Quix0r:
I use these scripts for easily restarting all my works (e.g. Errupter and Antminer have different hash rates).

They, of course, needs improvement as they may only fit my mining rig. :)


luke-jr:
Can you integrate this into start-bfgminer.sh in a not-tied-to-you way?

PR #434, opened 2014-04-22
Display an informational notice if the BFGMiner window is hidden

luke-jr:
This is to combat malware embedding BFGMiner.

### DIFF
--- a/miner.c
+++ b/miner.c
@@ -10397,6 +10397,45 @@ static void raise_fd_limits(void)
 #endif
 }
 
+#if defined(WIN32) && (defined(HAVE_OPENCL) || defined(WANT_CPUMINE))
+// If all devices are CPU/GPU, and the console window is hidden, show a message to confirm the user intends to be mining after a few seconds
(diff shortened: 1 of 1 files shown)


luke-jr:
@nwoolls Will this make problems for MultiMiner and/or other frontends?

nwoolls:
I'll need to test it but, if it does what it seems to do, yes this would effectively "break" MultiMiner in the eyes of users.

MultiMiner starts any number of BFGMiner instances to mine, depending on configuration. So if this displays on startup, that means a MultiMiner user pressing the "Start" button to start mining would see anywhere from 1, to 2, to N number of these prompts.

On top of that, one of the key features in MultiMiner is that it will restart BFGMiner unattended for any number of reasons, either due to crash or low Effective Hashrate etc. The upshot being that this "feature" of MultiMiner (restarting BFGMiner unattended) would be broken by dialogs being shown each time it happens.

luke-jr:
Seems this would be problematic, closing unmerged.

PR #445, opened 2014-05-10
gridseed: Switch from the Queued miner loop to the Scanhash miner loop

nwoolls:
Improves performance on Raspberry Pi
Fixes inaccurate hashrate estimates
Fixes bug overwriting work inadvertently

### DIFF
--- a/driver-gridseed.c
+++ b/driver-gridseed.c
@@ -20,6 +20,7 @@
 
 #define GRIDSEED_DEFAULT_FREQUENCY  600
 #define GRIDSEED_MAX_QUEUED          10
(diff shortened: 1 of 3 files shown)


driver-gridseed.c
```diff
@@ -38,7 +39,6 @@ struct cgpu_info *gridseed_alloc_device(const char *path, struct device_drv *dri
 	device->device_path = strdup(path);
 	device->device_fd = -1;
 	device->threads = 1;
-	device->procs = GC3355_ORB_DEFAULT_CHIPS;
```
luke-jr, line 41: This seems like a regression?

PR #451, opened 2014-05-21
gridseed: Persist the device serial number (dev_serial) during initializ...

nwoolls:
Persist the device serial number (dev_serial) during initialization


luke-jr:
Unnecessary.

PR #455, opened 2014-05-26
Bugfix: Homebrew: fix a segfault when BFGMiner is installed on Mac OS X

nwoolls:
Bugfix: Homebrew: fix a segfault when BFGMiner is installed on Mac OS X using Homebrew


luke-jr:
Merged

PR #456, opened 2014-05-29
Increase stratum proxy difficulty to 16 for SHA256 clients.

xrobau:
Difficulty of 1 is too little for even entry level 100gh hardware now.. And with 'entry level' moving to 500gh over the next few months, this should do for a while.

### DIFF
--- a/driver-stratum.c
+++ b/driver-stratum.c
@@ -343,7 +343,7 @@ void stratumsrv_mining_subscribe(struct bufferevent *bev, json_t *params, const 
 	if (opt_scrypt)
 		bufferevent_write(bev, "0.000015258556232", 17);
 	else
-		bufferevent_write(bev, "0.9999847412109375", 18);
+		bufferevent_write(bev, "15.999755859375", 15);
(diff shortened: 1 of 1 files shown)


luke-jr:
This doesn't check that the upstream difficulty isn't < 16. Maybe it should also be configurable?

xrobau:
That's deliberate.  Any pool that's asking for difficulty 1 shares with more than 100gh or so pointed at it is going to get flooded (difficulty 1 @ 100gh = about 30 shares per second). By filtering them to difficulty 16, it doesn't destroy the pool quite as badly, and gives it a chance to realise that it should increase the difficulty quickly.

I'm working on the assumption that people who are using the stratum proxy are using it for significant hashrates.  This may be wrong if people are using it with a couple of 1gh USB miners.  You know best! 8)

luke-jr:
That's the pool's problem - not sure we should make a bug to help them deal with it :)

luke-jr:
Please help test and review #486

PR #457, opened 2014-05-29
Increase usefulness of sharelog by reporting difficulty levels of shares

xrobau:
Add some more information to the sharelog script:

4997200,accept,ffffffffffffffffffffffffffffffffffffffffffffff4b3789410000000000,stratum+tcp://54.206.71.186:9332,PXY16,16,5f844a9c8d58a82f9c050993944a85ebe9d5d02469fdd490f04f2b0000000000,000000027ced44666ce97d508cc41ff9a824192de76a0af72142c9dc0000000000000000aca88457ed4606de7be505675a95e0d9f3f9c2b578333a110db387852a3254925387942018692842dcbd2626000000800000000000000000000000000000000000000000000000000000000000000000000000000000000080020000,1.51k/1k,1513.104908,1000.000000

Note the additional human-readable difficulty, plus the floats of the generated and requested difficulty. This is to ease post-processing of sharelogs later.

### DIFF
--- a/README
+++ b/README
@@ -644,16 +644,27 @@ To log share data to a file named "share.log", you can use either:
 
 For every share found, data will be logged in a CSV (Comma Separated Value)
 format:
-    timestamp,disposition,target,pool,dev,thr,sharehash,sharedata
+    timestamp,disposition,target,pool,dev,thr,sharehash,sharedata,workdiff/target,found,requested
 For example (this is wrapped, but it's all on one line for real):
-    1335313090,reject,
-    ffffffffffffffffffffffffffffffffffffffffffffffffffffffff00000000,
-    http://localhost:8337,GPU0,0,
-    6f983c918f3299b58febf95ec4d0c7094ed634bc13754553ec34fc3800000000,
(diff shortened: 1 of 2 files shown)


luke-jr:
What is the purpose of logging the "found diff"? It's a meaningless number...?

xrobau:
I added some documentation as to why.  Specifically, some devices have historically had issues submitting higher value shares (specifically, bitmain hardware has been known to be suspiciously unlucky).  It's a useless piece of information UNLESS you want it. In which case it's invaluable, and much easier than calculating it from the sharelog itself.

luke-jr:
Hmm. It's not sufficient to just put the raw float values? :)

xrobau:
The raw float values are there. (last two fields).

luke-jr:
Hence "just"

xrobau:
As I said on IRC, they're there to ease post-processing.  A simple grep/cut/sort is enough to get all the useful information from the file with this patch. Otherwise a complete parser would have to be written.

PR #460, opened 2014-06-02
gridseed: added support for the 80-chip G-Blade Scrypt-only miner

nwoolls:
gridseed: added support for the 80-chip (two blades of 40 chips) G-Blade Scrypt-only miner

### DIFF
--- a/gc3355.c
+++ b/gc3355.c
@@ -737,7 +737,7 @@ uint32_t gc3355_get_firmware_version(int fd)
 		return -1;
 	}
 	
-	uint32_t fw_version = le32toh(*(uint32_t *)(buf + 8));
+	uint32_t fw_version = be32toh(*(uint32_t *)(buf + 8));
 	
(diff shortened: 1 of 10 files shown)


gc3355.c
```diff
@@ -737,7 +737,7 @@ uint32_t gc3355_get_firmware_version(int fd)
 		return -1;
 	}
 	
-	uint32_t fw_version = le32toh(*(uint32_t *)(buf + 8));
+	uint32_t fw_version = htobe32(*(uint32_t *)(buf + 8));
```
luke-jr: This looks like it should be be32toh... And changes the meaning from previous code - is other use okay with it?

nwoolls: Ah that makes sense and works well. As far as changing the result you are correct, but was never really used other than checking for -1

luke-jr: uint32_t can never be -1 O.o

PR #461, opened 2014-06-03
Bugfix: gridseed: use a signed int so returning -1 has defined behavior

nwoolls:


luke-jr:
Will int32 have enough values to cover the non-negative1 range?

driver-gridseed.c
```diff
@@ -98,7 +98,7 @@ bool gridseed_detect_custom(const char *path, struct device_drv *driver, struct
 	gc3355_init_usborb(device->device_fd, info->freq, false, false);
 	
 	applog(LOG_INFO, "Found %"PRIpreprv" at %s", device->proc_repr, path);
-	applog(LOG_DEBUG, "%"PRIpreprv": Init: firmware=%d", device->proc_repr, fw_version);
+	applog(LOG_DEBUG, "%"PRIpreprv": Init: firmware=%lld", device->proc_repr, fw_version);
```
luke-jr: fw_version needs to be cast to (long long) here.

PR #464, opened 2014-06-03
gridseed: Allow specifying arbitrary number of chips with --set gsd:chips=X

nwoolls:


driver-gridseed.c
```diff
@@ -255,11 +255,11 @@ int64_t gridseed_scanhash(struct thr_info *thr, struct work *work, int64_t __may
 static
 char *gridseed_set_device(struct cgpu_info *device, char *option, char *setting, char *replybuf)
 {
+	int val = atoi(setting);
+	struct gc3355_orb_info *info = (struct gc3355_orb_info *)(device->device_data);
```
luke-jr: Double-cast is unnecessary.

luke-jr:
Doesn't have to be now, but at some point it would be nice to upgrade gridseed to the set_device_funcs interface, as I'd like to obsolete .set_device at some point.

luke-jr:
I think the logged chip count will still be wrong, but meh.

PR #471, opened 2014-06-20
zeusminer: support added for ZeusMiner scrypt ASICs

nwoolls:
--set zus:freq=XYZ (default 328)
--set zus:chips=X (default 6)
--set zus:cores=X (default 8)
--set zus:ignore_golden_nonce (don't check probe response - buggy HW/FW)

### DIFF
--- a/driver-icarus.c
+++ b/driver-icarus.c
@@ -413,27 +413,7 @@ const char *icarus_set_timing(struct cgpu_info * const proc, const char * const 
 
 static uint32_t mask(int work_division)
 {
-	uint32_t nonce_mask = 0x7fffffff;
-
-	// yes we can calculate these, but this way it's easy to see what they are
-	switch (work_division) {
-	case 1:
-		nonce_mask = 0xffffffff;
-		break;
-	case 2:
-		nonce_mask = 0x7fffffff;
-		break;
-	case 4:
-		nonce_mask = 0x3fffffff;
-		break;
-	case 8:
-		nonce_mask = 0x1fffffff;
-		break;
-	default:
-		quit(1, "Invalid2 work_division (%d) must be 1, 2, 4 or 8", work_division);
-	}
-
-	return nonce_mask;
+	return 0xffffffff / upper_power_of_two(work_division);
 }
 
 // Number of bytes remaining after reading a nonce from Icarus
--- a/driver-icarus.c
+++ b/driver-icarus.c
@@ -488,45 +488,48 @@ bool icarus_detect_custom(const char *devpath, struct device_drv *api, struct IC
 
 	if (info->detect_init_func)
 		info->detect_init_func(devpath, fd, info);
-	
-	int ob_size = strlen(info->golden_ob) / 2;
-	unsigned char ob_bin[ob_size];
-	BFGINIT(info->ob_size, ob_size);
-	
-	hex2bin(ob_bin, info->golden_ob, sizeof(ob_bin));
-	icarus_write(fd, ob_bin, sizeof(ob_bin));
-	cgtime(&tv_start);
 
-	memset(nonce_bin, 0, sizeof(nonce_bin));
-	// Do not use info->read_size here, instead read exactly ICARUS_NONCE_SIZE
-	// We will then compare the bytes left in fd with info->read_size to determine
-	// if this is a valid device
-	icarus_gets(nonce_bin, fd, &tv_finish, NULL, info->probe_read_count, ICARUS_NONCE_SIZE);
-	
-	// How many bytes were left after reading the above nonce
-	int bytes_left = icarus_excess_nonce_size(fd, info);
-	
-	icarus_close(fd);
+	if (!info->ignore_golden_nonce)
+	{
+		int ob_size = strlen(info->golden_ob) / 2;
+		unsigned char ob_bin[ob_size];
+		BFGINIT(info->ob_size, ob_size);
 
-	bin2hex(nonce_hex, nonce_bin, sizeof(nonce_bin));
-	if (!info->ignore_golden_nonce && strncmp(nonce_hex, info->golden_nonce, 8))
-	{
-		applog(LOG_DEBUG,
-			"%s: "
-			"Test failed at %s: get %s, should: %s",
-			api->dname,
-			devpath, nonce_hex, info->golden_nonce);
-		return false;
-	}
-		
-	if (info->read_size - ICARUS_NONCE_SIZE != bytes_left) 
-	{
-		applog(LOG_DEBUG,
-			   "%s: "
-			   "Test failed at %s: expected %d bytes, got %d",
-			   api->dname,
-			   devpath, info->read_size, ICARUS_NONCE_SIZE + bytes_left);
-		return false;
+		hex2bin(ob_bin, info->golden_ob, sizeof(ob_bin));
+		icarus_write(fd, ob_bin, sizeof(ob_bin));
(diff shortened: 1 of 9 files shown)


driver-icarus.c
```diff
@@ -529,7 +529,7 @@ bool icarus_detect_custom(const char *devpath, struct device_drv *api, struct IC
 	icarus_close(fd);
 
 	bin2hex(nonce_hex, nonce_bin, sizeof(nonce_bin));
-	if (strncmp(nonce_hex, info->golden_nonce, 8))
+	if (!info->ignore_golden_nonce && strncmp(nonce_hex, info->golden_nonce, 8))
```
luke-jr: Is there a reason to even send/receive the golden packets if the result is ignored?

driver-icarus.c
```diff
@@ -1068,9 +1068,15 @@ static int64_t icarus_scanhash(struct thr_info *thr, struct work *work,
 		inc_hw_errors(thr, state->last_work, nonce);
 	icarus_transition_work(state, work);
 
-	hash_count = (nonce & info->nonce_mask);
-	hash_count++;
-	hash_count *= info->fpga_count;
+	if (info->ignore_nonce_mask)
+		hash_count = ((double)(elapsed.tv_sec)
+					  + ((double)(elapsed.tv_usec))/((double)1000000)) / info->Hs;
```
luke-jr: What's the logic here? The nonce_mask exists because it's needed...

driver-icarus.h
```diff
@@ -110,15 +126,24 @@ struct ICARUS_INFO {
 	
 	// Bytes to read from Icarus for nonce
 	int read_size;
-	
+
```
luke-jr: Please don't break the indentation :)

driver-zeusminer.c
```diff
@@ -0,0 +1,296 @@
...
+static
+uint32_t zeusminer_calc_clk_header(uint16_t freq)
+{
+	int chip_clk = freq;
+
+	//max clock 383MHz, min clock 200MHz
+	if (chip_clk > 383)
+		chip_clk = 383;
+	else if (chip_clk < 200)
+		chip_clk = 200;
```
luke-jr: Should error when setting invalid values, NOT check them here.

driver-zeusminer.c
```diff
@@ -0,0 +1,296 @@
...
+bool zeusminer_job_prepare(struct thr_info *thr, struct work *work, __maybe_unused uint64_t max_nonce)
+{
+	struct cgpu_info * const device = thr->cgpu;
+	struct icarus_state * const state = thr->cgpu_data;
+	struct ICARUS_INFO * const info = device->device_data;
+
+	memset(state->ob_bin, 0, info->ob_size);
+
+	uint32_t clk_header = zeusminer_calc_clk_header(info->freq);
+	uint32_t diff = floor(zeusminer_calc_target_diff(work->target));
```
luke-jr: Hardware should never use work->target, just work->nonce_diff.

driver-zeusminer.c
```diff
@@ -0,0 +1,296 @@
...
+	d64 = be64toh(*data64);
+
+	if(unlikely(!d64))
+		d64 = 1;
+
+	return zeusminer_diff_one / d64;
+}
+
+static
+void zeusminer_reverse_bytes(unsigned char *s, size_t l)
```
luke-jr: Doesn't driver-icarus already have something like this?

util.c
```diff
@@ -3524,3 +3524,16 @@ uint8_t crc8ccitt(const void * const buf, const size_t buflen)
 		crc = _crc8ccitt_table[crc ^ *p++];
 	return crc;
 }
+
+// http://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
+uint32_t nearest_pow(uint32_t num)
```
luke-jr: This seems to round up, not nearest?

driver-icarus.c
nwoolls: It's not all of them, only a small portion. Most work fine and respond to the golden OB. Some do not, but still hash fine if you tell them to get on with it. Verified myself with user HW.

driver-icarus.c
nwoolls: I've gone through the Icarus code and determined:
- work_division is used to set nonce_mask
- it is also used to set fpga_count if fpga_count is not set
- fpga_count and nonce_mask are only used to calculate hashes when a nonce is found

Based on all of my testing (and other users / forks), it's not possible to get the nonce_mask / work_division combo to work with e.g. 60 chips and 8 cores each. The nonce_mask code is meant to take a power of 2 only.

driver-icarus.c
luke-jr: Right, my point is why bother sending and waiting for the response, when you're just going to ignore it? Why not just skip that and add_cgpu etc (when this flag is set)?

driver-icarus.h
nwoolls: I don't think this broke it - I think this actually removed unnecessary indentation / whitespace.

util.c
nwoolls: It does - I actually did this to maintain the name with a previous (wrong) implementation. I'll change it since it's a clean commit now.

driver-icarus.c
luke-jr: I don't understand the problem. How is it dividing up work? Wouldn't just changing & to % work?

driver-icarus.h
luke-jr: That's what's broken - it should be indented like the rest of the block :)

driver-zeusminer.c
nwoolls: Can you advise on how to incorporate zeusminer_calc_target_diff() based on using work->nonce_diff instead? This was a fix that was required by jstefanop in the Darkwinde fork to get the ASIC to work with some difficulties.

floor(zeusminer_calc_target_diff(work->target)) is 255.00 for me

luke-jr:
```
diff = work->nonce_diff * 0x10000;
```

Note that min_nonce_diff is set such that this can never be <1

driver-zeusminer.c
jstefanop: no this is specifically for scrypt algo

driver-zeusminer.c
jstefanop: nonce_diff is not calculated correctly for scrypt in miner.c...I tried fixing this but this causes a mess of issues because of how the nonce is validated, hash rates displayed, etc, so was much simpler to calculate the diff from target directly.

jstefanop:
BTW all Zeus gen 1 chips have 8 cores and will probably be that way for a while...might not want to have the core # user configurable

driver-zeusminer.c
nwoolls: Can you test the latest code in my -clean branch to ensure it is broken by making Luke's suggested changes? It works well for me but I know you have more experience with this bug.

nwoolls:
Sorry folks - my Push hid some comments. @jstefanop please test this latest code if you can and verify if your diff fixes are still required or if Luke's suggestions work.

jstefanop:
Diff changes and your last commits look good!

nwoolls:
Score! Thanks so much for confirming.

luke-jr:
@jstefanop Can you test the zeusminer branch on my github repo?

nwoolls:
SSH in if you need to Luke. There's a Zeus plugged in. 

-----Original Message-----

PR #473, opened 2014-06-20
zeusminer: support added for ZeusMiner scrypt ASICs

nwoolls:
--set zus:freq=XYZ (default 328)
--set zus:chips=X (default 6)
--set zus:cores=X (default 8)
--set zus:ignore_golden_nonce (don't check probe response - buggy HW/FW)

### DIFF
--- a/driver-icarus.c
+++ b/driver-icarus.c
@@ -512,42 +512,48 @@ bool icarus_detect_custom(const char *devpath, struct device_drv *api, struct IC
 	int ob_size = strlen(info->golden_ob) / 2;
 	unsigned char ob_bin[ob_size];
 	BFGINIT(info->ob_size, ob_size);
-	
-	hex2bin(ob_bin, info->golden_ob, sizeof(ob_bin));
-	icarus_write(fd, ob_bin, sizeof(ob_bin));
-	cgtime(&tv_start);
 
-	memset(nonce_bin, 0, sizeof(nonce_bin));
-	// Do not use info->read_size here, instead read exactly ICARUS_NONCE_SIZE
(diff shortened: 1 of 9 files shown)


driver-icarus.c
```diff
@@ -413,27 +413,7 @@ const char *icarus_set_timing(struct cgpu_info * const proc, const char * const
...
-		break;
-	case 8:
-		nonce_mask = 0x1fffffff;
-		break;
-	default:
-		quit(1, "Invalid2 work_division (%d) must be 1, 2, 4 or 8", work_division);
-	}
-
-	return nonce_mask;
+	return 0xffffffff / upper_power_of_two(work_division);
```
luke-jr, line 416: Seems better to have a sanity check and error if it isn't a power-of-two value?

driver-zeusminer.c
```diff
@@ -0,0 +1,252 @@
...
+	struct cgpu_info * const device = thr->cgpu;
+	struct icarus_state * const state = thr->cgpu_data;
+	struct ICARUS_INFO * const info = device->device_data;
+
+	uint32_t clk_header = zeusminer_calc_clk_header(info->freq);
+	uint32_t diff = work->nonce_diff * 0x10000;
+	uint32_t target_me = 0xffff / diff;
+	uint32_t header = clk_header + target_me;
+
+	swabn(state->ob_bin, (uint8_t *)&header, 4);
```
luke-jr, line 217: This is undefined C and _will_ break on some endian. Use pk_u32le or pk_u32be

luke-jr:
Merged

PR #476, opened 2014-06-27
icarus: Bugfix: don't use invalid nonces for calculating hash_count

nwoolls:


driver-icarus.c
```diff
@@ -1063,14 +1063,22 @@ static int64_t icarus_scanhash(struct thr_info *thr, struct work *work,
...
 	else
+	{
 		inc_hw_errors(thr, state->last_work, nonce);
-	icarus_transition_work(state, work);
 
-	hash_count = (nonce & info->nonce_mask);
-	hash_count++;
-	hash_count *= info->fpga_count;
+		// Estimate hashrate based on info->Hs, nonce was invalid
+		hash_count = ((double)(elapsed.tv_sec) + ((double)(elapsed.tv_usec))/((double)1000000)) / info->Hs;
```
luke-jr, line 1079: Please share code with the idle estimation. Also need to move the debug message into the conditional...

PR #481, opened 2014-07-01
Bugfix: probe for ZeusMiner before probing for DualMiner

nwoolls:
Without this, -S all with a DualMiner conflicts with a ZeusMiner


luke-jr:
"Conflicts" how?

nwoolls:
Causes the Zus not to respond to its own probe

luke-jr:
Won't the Zus probe cause the DualMiner to not respond? :/

nwoolls:
Nope - I tried the various combinations. The only issue is sending the DM probe to the ZM.

PR #488, opened 2014-07-06
rockminer: implement --device-protocol-dump for debugging

nwoolls:


luke-jr:
I wonder if these should just be in lowl-vcom somehow :)

nwoolls:
I was thinking the same. The same code getting used over and over. It's minor code but still.

PR #489, opened 2014-07-07
lowl-vcom: Add support for auto scanning (-S auto) on Mac OS X

nwoolls:
Uses I/O Kit to enumerate USB devices
https://developer.apple.com/library/mac/documentation/DeviceDrivers/Conceptual/IOKitFundamentals/Introduction/Introduction.html
http://en.wikipedia.org/wiki/I/O_Kit

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -182,6 +182,7 @@ case $target in
     use_udevrules=false
     ;;
 	*-*-darwin*)
+		LDFLAGS="$LDFLAGS -framework IOKit -framework CoreFoundation"
 		have_macho=true
     use_udevrules=false
 		;;
--- a/lowl-vcom.c
+++ b/lowl-vcom.c
@@ -3,6 +3,7 @@
  * Copyright 2013 Con Kolivas
  * Copyright 2012 Andrew Smith
  * Copyright 2013 Xiangfu
+ * Copyright 2014 Nate Woolls
  *
  * This program is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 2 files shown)


configure.ac
```diff
@@ -182,6 +182,7 @@ case $target in
     use_udevrules=false
     ;;
 	*-*-darwin*)
+		LDFLAGS="$LDFLAGS -framework IOKit -framework CoreFoundation"
```
luke-jr, line 185: Is IOKit available on all versions of Darwin?

lowl-vcom.c
```diff
@@ -187,6 +194,127 @@ void _vcom_devinfo_scan_udev(struct lowlevel_device_info ** const devinfo_list)
...
+
+	kern_return_t kret = (*usb_device)->DeviceRequest(usb_device, &dev_req);
+	if (kret != 0)
+		return NULL;
+
+	char * result = malloc(128);
+	size_t str_len = (dev_req.wLenDone - 1) / 2;
+	uint32_t i;
+	for (i = 0; i < str_len; i++)
+		result[i] = buf[i + 1];
```
luke-jr: ucs2_to_utf8_dup?

configure.ac
nwoolls, line 185: Yes. Darwin is just the OSS kernel for OS X (XNU). I/O Kit is the I/O framework for the XNU kernel in OS X and NeXTSTEP.

PR #497, opened 2014-07-11
gridseed: Support devices with more than 255 physical processors

nwoolls:


luke-jr:
I'm confused. Didn't we already do this? O.o

nwoolls:
That was for ZM

PR #498, opened 2014-07-11
gridseed: Add support for displaying per-chip statistics (press D, M, Return)

nwoolls:


luke-jr:
I believe this breaks changing clock via RPC. Please define gridseed_set_device_funcs_probe (for drv_set_defaults, for probe-time settings) separate from gridseed_set_device_funcs (for RPC pgaset, at runtime).

nwoolls:
Let me know how this looks.

luke-jr:
No better? I don't see any changes to the issue mentioned... Probably would be cleared if you put the set_device_funcs refactor in a separate commit first?

nwoolls:
What specifically is broken by this PR. Please comment on a specific line.

luke-jr:
It's about missing lines :)

nwoolls:
This should be addressed now and ready for merge. I've tested setting clock both during probe and via RPC, as well as setting chips via probe.

luke-jr:
Merged as a253680a9a22bd6d35d71e1de2fa950204d3ab17, ebef06a72a519b0ec289edd7806f41a1e6dfe935, and 17636b2211b62b47ddb358f9a51c6e2568a54c3c

PR #500, opened 2014-07-11
zeusminer: Add support for displaying per-chip statistics (press D, M, Return)

nwoolls:


driver-icarus.c
```diff
@@ -551,6 +551,11 @@ bool icarus_detect_custom(const char *devpath, struct device_drv *api, struct IC
 	icarus->device_fd = -1;
 	icarus->threads = 1;
 	icarus->set_device_funcs = icarus_set_device_funcs;
+
+#ifdef USE_ZEUSMINER
+	icarus->procs = info->chips;
```
luke-jr, line 556: Why not enable this for all icarus-based devices? Should be procs = fpga_count...

nwoolls, line 556: For ZUS fpga_count = chips \* cores, which would not be the correct number of procs and would not have the correct information submitted later by finding the thread for the nonce.

**Edit**: confirmed this doesn't work - only 6 of the 48 "processors" show shares.

**Edit #2**: left it running for longer and the other "processors" do indeed seem to be showing shares. The only issue I see is whether folks would want to see, e.g., 256 \* 8 individual devices (vs. 256). Thoughts?

**Edit #3**: I would think seeing stats by core rather than by chip would be _less_ helpful. It is going to be harder to ID problem chips, no?

luke-jr, line 556: Nah, it's just a matter of division or, if you add the info, procdetails to see which chip. IMO as a rule, more information is better.

nwoolls:
Why is this closed?

driver-icarus.c
nwoolls, line 556: Please explain in detail. A matter if division how? The user would need to either count the rows displayed or use the heading (0ij) and try to figure out the chip / core from that.

I really do not, at all, see how per-core statistics are beneficial over per-chip statistics. Can you give a concrete example (rather than a general rule)?

jstefanop:
Yea per core would make no sense, would rather see per chips stats

nwoolls:
Closing this to work on another PR that builds off the Icarus per-proc stats

nwoolls:
@jstefanop I lost this battle - the stats are per-core not per-chip. I'm going to try to get the chips identified visually in the UI though as I agree, visually speaking this is less helpful.

luke-jr:
The precedent had already been set for the current interfaces, and per-core can always be reduced to per-chip by programs using RPC etc. I agree it makes sense to give per-chip (or per-board) summaries in general, but this should be a generic solution, not one driver standing out as an oddball. :)

nwoolls:
@luke-jr same page - I think this is a better technical solution but leaves more lacking for the end-user (for now).

nwoolls:
@luke-jr is procdetails API only? I would like something in the UI

nwoolls:
@jstefanop I'm expanding on this work on the Icarus driver to include Chip # in the ZeusMiner driver (both UI and API):

ZeusMiner Stats

Issue #502, opened 2014-07-12
gen-version.sh: Syntax error: redirection unexpected


gades:
~/src/bfgminer# ./autogen.sh 
Getting submodules...
Submodule 'ccan' (git://git.ozlabs.org/~ccan/ccan) registered for path 'ccan-upstream'
Submodule 'libblkmaker' (git://gitorious.org/bitcoin/libblkmaker.git) registered for path 'libblkmaker'
Cloning into 'ccan-upstream'...
remote: Counting objects: 16194, done.
remote: Compressing objects: 100% (4746/4746), done.
remote: Total 16194 (delta 11759), reused 15437 (delta 11227)
Receiving objects: 100% (16194/16194), 12.60 MiB | 1.93 MiB/s, done.
Resolving deltas: 100% (11759/11759), done.
Checking connectivity... готово.
Submodule path 'ccan-upstream': checked out '8b0bdb090e2882aa431e89f4bc7aa4736e9e2838'
Cloning into 'libblkmaker'...
remote: Counting objects: 495, done.
remote: Compressing objects: 100% (491/491), done.
remote: Total 495 (delta 304), reused 0 (delta 0)
Receiving objects: 100% (495/495), 103.51 KiB | 0 bytes/s, done.
Resolving deltas: 100% (304/304), done.
Checking connectivity... готово.
Submodule path 'libblkmaker': checked out '6cc0258704ece616f2de6e5509abfc7fba108957'
Running autoreconf -if...
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, `.'.
libtoolize: copying file`./ltmain.sh'
libtoolize: Consider adding `AC_CONFIG_MACRO_DIR([m4])' to configure.ac and
libtoolize: rerunning libtoolize, to keep the correct libtool macros in-tree.
libtoolize: Consider adding`-I m4' to ACLOCAL_AMFLAGS in Makefile.am.
configure.ac:18: installing './ar-lib'
configure.ac:17: installing './compile'
configure.ac:19: installing './config.guess'
configure.ac:19: installing './config.sub'
configure.ac:14: installing './install-sh'
configure.ac:14: installing './missing'
Makefile.am: installing './depcomp'
configure.ac:43: installing './compile'
configure.ac:30: installing './config.guess'
configure.ac:30: installing './config.sub'
configure.ac:41: installing './install-sh'
configure.ac:41: installing './missing'
[...]

PR #503, opened 2014-07-12
gen-version.sh: Syntax error: redirection unexpected

gades:
Fix issue #502


luke-jr:
Merged as 54377f3, thanks!

PR #507, opened 2014-07-12
zeusminer: Identify the Chip # both in the UI and RPC API

nwoolls:
ZeusMiner Stats

### DIFF
--- a/driver-zeusminer.c
+++ b/driver-zeusminer.c
@@ -229,6 +229,18 @@ bool zeusminer_job_prepare(struct thr_info *thr, struct work *work, __maybe_unus
 	return true;
 }
 
+// display the Chip # in the UI when viewing per-proc details
+static
+bool zeusminer_override_statline_temp2(char *buf, size_t bufsz, struct cgpu_info *device, __maybe_unused bool per_processor)
+{
+	if (per_processor && ((device->proc_id % ZEUSMINER_CHIP_CORES) == 0))
+	{
+		tailsprintf(buf, bufsz, "C:%-3d", device->proc_id / ZEUSMINER_CHIP_CORES);
+		return true;
+	}
+	return false;
+}
+
 // device_drv definition - miner.h
 
 static
@@ -254,6 +266,10 @@ void zeusminer_drv_init()
 	// specify driver probe priority
 	// currently setup specifically to probe before DualMiner
 	zeusminer_drv.probe_priority = -100;
+
+	// output the chip # when viewing per-proc stats
+	// so we can easily ID chips vs cores
+	zeusminer_drv.override_statline_temp2 = zeusminer_override_statline_temp2;
 }
 
 struct device_drv zeusminer_drv = {
--- a/driver-zeusminer.c
+++ b/driver-zeusminer.c
@@ -241,6 +241,14 @@ bool zeusminer_override_statline_temp2(char *buf, size_t bufsz, struct cgpu_info
 	return false;
 }
 
+// return the Chip # in via the API when procdetails is called
+static
+struct api_data *zeusminer_get_api_extra_device_detail(struct cgpu_info *device)
+{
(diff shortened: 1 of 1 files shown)


driver-zeusminer.c
```diff
@@ -217,6 +217,30 @@ bool zeusminer_job_prepare(struct thr_info *thr, struct work *work, __maybe_unus
...
+	return false;
+}
+
+// return the Chip # in via the API when procdetails is called
+static
+struct api_data *zeusminer_get_api_extra_device_detail(struct cgpu_info *device)
+{
+	// only add to procdetails, not devdetails
+	if (device->device == device)
+		return NULL;
```
luke-jr: devdetails doesn't call this at all. (device->device == device) simply means it's the first proc...
(also suggest renaming the var to proc)

driver-zeusminer.c
```diff
@@ -217,6 +217,30 @@ bool zeusminer_job_prepare(struct thr_info *thr, struct work *work, __maybe_unus
 	return true;
 }
 
+// display the Chip # in the UI when viewing per-proc details
+static
+bool zeusminer_override_statline_temp2(char *buf, size_t bufsz, struct cgpu_info *device, __maybe_unused bool per_processor)
+{
+	if (per_processor && ((device->proc_id % ZEUSMINER_CHIP_CORES) == 0))
```
luke-jr, line 236: Might want to remove the 2nd condition here, since zeusminer_override_statline_temp2 is used in more than just the device list (it's used in Manage TUI)

driver-zeusminer.c
nwoolls: devdetails definitely calls this. That's why I added the check. Without the check devdetails adds a chip=0 to the end.

nwoolls: See the following:

https://github.com/luke-jr/bfgminer/blob/bfgminer/api.c#L3412
->
https://github.com/luke-jr/bfgminer/blob/bfgminer/api.c#L1608
->
https://github.com/luke-jr/bfgminer/blob/bfgminer/api.c#L1608
->
https://github.com/luke-jr/bfgminer/blob/bfgminer/api.c#L1598

Further, that's where the checks for cgpu->device == cgpu came from:

https://github.com/luke-jr/bfgminer/blob/bfgminer/api.c#L1598
https://github.com/luke-jr/bfgminer/blob/bfgminer/api.c#L2909

Is there a different check I can do to determine if get_api_extra_device_detail is per-proc or not?

driver-zeusminer.c
nwoolls, line 236: What problem does this cause for the Manage TUI? It looks / works correctly there.

The Chip # should not be on every single core, that's just overkill. If there's a downside I'm not seeing let me know.

driver-zeusminer.c
luke-jr: https://github.com/luke-jr/bfgminer/blob/bfgminer/api.c#L1544 appears to be missing for the devdetails case. I'll fix that...

driver-zeusminer.c
luke-jr, line 236: The summary shown in the content for Manage won't have the chip number except for the first chip. If that's okay with you, it's okay with me.

nwoolls, line 236: I am not seeing that. I see all chips IDd in the Management UI (hitting M in the TUI).

It seems to work in the Management UI just like it does when viewing per-proc stats (D, M, Return).

driver-zeusminer.c
nwoolls: Let me know when that is fixed and merged and I'll re-base my changes on that. I should then be able to remove the above lines.

luke-jr: Already did a few days ago.

nwoolls:
Updated this to remove the (now) unnecessary check for devdetails vs procdetails.

PR #513, opened 2014-07-14
gc3355: Re-initialize gc3355-based ASICs when triggered by the watchdog

nwoolls:


nwoolls:
Closing pending additional testing.

PR #516, opened 2014-07-16
util: Work around broken Scrypt pools using realistic pdiff assumptions

nwoolls:
Reasoning: we need to add another condition to the existing "if (opt_scrypt)" work-around as ASICs approach pdiff 1.
If we are to add an additional check for pdiff > N, it makes sense to use a realistic value to adjust all Scrypt pool diffs.

### DIFF
--- a/util.c
+++ b/util.c
@@ -2460,22 +2460,34 @@ static bool parse_diff(struct pool *pool, json_t *val)
 
 	if ((int64_t)diff != diff)
 	{
-		// Always assume fractional values are proper bdiff per specification
+		// Assume fractional values are proper bdiff per specification
+		// Allow integers to be interpreted as pdiff, since the difference is trivial and some pools see it this way
 		diff = bdiff_to_pdiff(diff);
-	}
-	else
-	{
-		// Integer; allow it to be interpreted as pdiff, since some the difference is trivial and some pools see it this way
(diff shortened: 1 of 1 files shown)


util.c
```diff
@@ -2460,22 +2460,34 @@ static bool parse_diff(struct pool *pool, json_t *val)
...
+
+	// Ideally pools will fix their implementation and we can remove this
+	// This should suffice until miners are hashing Scrypt at ~4-32 Gh/s (based on a share rate target of 10-60s)
+
+	const double minimum_broken_scrypt_diff = 64;
+	// Diff 64 at 4.26 Gh/s = 1 share / 60s
+	// Diff 64 at 25.0 Gh/s = 1 share / 13s
+	// Diff 64 at 32.0 Gh/s = 1 share / 10s
+
+	if (opt_scrypt && (diff > minimum_broken_scrypt_diff))
```
luke-jr: Not >= ?

nwoolls: Ya not >=, I should rename the const to be minimum_acceptable_scrypt_diff, the intent was to allow 64 to be a "real" Scrypt diff. See the comments above. I've been unable to find a pool that has anything lower than 128 for even CPU miners.

nwoolls:
Addressed your logic question in the latest push.

PR #520, opened 2014-07-21
Bugfix: zeusminer: icarus timing can lead to false positive Idle detection

nwoolls:


luke-jr:
Huh? Why is that?

nwoolls:
It causes the read_count to reach a point where the thread doesn't report in for 60 seconds, so BFGMiner reports it as idle.

luke-jr:
Hm, should it be > 60 seconds? Maybe we should just have icarus reportin?

luke-jr:
Actually, we should check scantime in icarus at least, and limit read_count to that too...

PR #524, opened 2014-07-24
Bugfix: util: Lower the threshold for considering pool Scrypt diff broken

nwoolls:


util.c
```diff
@@ -2476,17 +2476,16 @@ static bool parse_diff(struct pool *pool, json_t *val)
 
 	/* 7/12/2014: P2Pool code was fixed: https://github.com/forrestv/p2pool/pull/210
 	   7/15/2014: Popular pools unfixed: wemineltc, dogehouse, p2pool.org
-                  Cannot find a broken Scrypt pool that will dispense diff 64 or lower */
+                  Cannot find a broken Scrypt pool that will dispense diff lower than 16 */
 
 	// Ideally pools will fix their implementation and we can remove this
-	// This should suffice until miners are hashing Scrypt at ~4-32 Gh/s (based on a share rate target of 10-60s)
+	// This should suffice until miners are hashing Scrypt at ~1-7 Gh/s (based on a share rate target of 10-60s)
```
luke-jr, line 2482: Aren't some miners awfully close to or even past 1-7 Gh/s already? Better to break with broken pools, than to break with working pools...

nwoolls, line 2482: I have a task out to add the #brokendiff and #nobrokendiff URIs. The current code in Git is broken for low diff Scrypt servers. The code prior to that is broken for Scrypt diffs (legit) greater than or equal to 1.

So I think this is a decent fix for what was broken, also works around broken P2Pool, and lets us push out a _more compatible_ build before putting out one that breaks compatibility in favor of being truly _right_.

nwoolls, line 2482: I'm seeing 88 Mh/s available from resellers currently.

PR #528, opened 2014-07-25
Bugfix: gridseed: Implement a proper scanhash routine for GridSeeds

nwoolls:
Previous behavior worked, but not intentionally

### DIFF
--- a/driver-gridseed.c
+++ b/driver-gridseed.c
@@ -20,7 +20,9 @@
 
 #define GRIDSEED_DEFAULT_FREQUENCY  600
 // 60Kh/s at 700MHz in ms
-#define GRIDSEED_HASH_SPEED			0.08571428571429
+#define GRIDSEED_HASH_SPEED         0.08571428571429
+// GridSeed driver currently scans a full nonce range
+#define GRIDSEED_MAX_NONCE          0xffffffff
 
 BFG_REGISTER_DRIVER(gridseed_drv)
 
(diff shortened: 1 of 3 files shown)


driver-gridseed.c
```diff
@@ -246,16 +259,21 @@ int64_t gridseed_scanhash(struct thr_info *thr, struct work *work, int64_t __may
 	unsigned char buf[GC3355_READ_SIZE];
 	int read = 0;
 	int fd = device->device_fd;
+	int64_t total_hashes = 0;
 
-	while (!thr->work_restart && (read = gc3355_read(fd, (char *)buf, GC3355_READ_SIZE)) > 0)
+	while (!thr->work_restart                                                   // true when new work is available (miner.c)
+	    && ((read = gc3355_read(fd, (char *)buf, GC3355_READ_SIZE)) >= 0)       // only check for failure - allow 0 bytes
+	    && (total_hashes < max_nonce))                                          // false when we've had time to scan a range
```
luke-jr: This isn't how it's supposed to work... max_nonces means you're _actually_ stopping at that point.

nwoolls: _Actually_ stopping as opposed to?

luke-jr: As opposed to the device continuing to hash the next nonce...

nwoolls: If the device has hashed more than max_nonce, it exits the loop, exits driver->scanhash, which then sends the next piece of work.

nwoolls:
Adjusted code to reference work->blk.nonce instead of max_nonce. Also added an applog call if a nonce range is scanned.

nwoolls:
Introduced GRIDSEED_MAX_NONCE and checking total_hashes vs 0xffffffff

driver-gridseed.c
```diff
@@ -90,13 +92,23 @@ int64_t gridseed_calculate_chip_hashes(struct thr_info *thr)
...
-	for (struct cgpu_info *proc = device; proc; proc = proc->next_proc)
+	for_each_managed_proc(proc, device)
+	{
+		total_hashes += chip_hashes;
 		hashes_done2(proc->thr[0], chip_hashes, NULL);
+	}
+
+	// not possible to complete more than 0xffffffff nonces
+	if (unlikely(total_hashes > 0xffffffff))
+		total_hashes = 0xffffffff;
```
luke-jr: This should consider hashes performed BEFORE this function call, on the same work.

nwoolls: Good call - made the change too fast. Meant it to be in gridseed_calculate_chip_hashes.

That said, gridseed_calculate_chip_hashes is doing per-chip hashes so this should be just a sanity check.

nwoolls:
Take a look again when you get a chance please.

driver-gridseed.c
```diff
@@ -76,27 +78,72 @@ struct thr_info *gridseed_thread_by_chip(const struct cgpu_info * const device,
...
+	int64_t previous_chip_hashes = previous_hashes / device->procs;
+    int64_t recent_chip_hashes = total_chip_hashes - previous_chip_hashes;
+	int64_t total_hashes = 0;
+
+	for_each_managed_proc(proc, device)
+	{
+		total_hashes += recent_chip_hashes;
+		hashes_done2(proc->thr[0], recent_chip_hashes, NULL);
+	}
+	total_hashes = gridseed_fix_hashes_done(total_hashes);
```
luke-jr: This should be redundant.

driver-gridseed.c
```diff
@@ -76,27 +78,72 @@ struct thr_info *gridseed_thread_by_chip(const struct cgpu_info * const device,
...
+// report on hashes done since start_tv
+// return the number of hashes done since start_tv
+static
+int64_t gridseed_hashes_done(struct cgpu_info * const device, struct timeval const start_tv, int64_t previous_hashes)
+{
+	int64_t total_chip_hashes = gridseed_calculate_chip_hashes(device, start_tv);
+	total_chip_hashes = gridseed_fix_hashes_done(total_chip_hashes);
+
+	int64_t previous_chip_hashes = previous_hashes / device->procs;
+    int64_t recent_chip_hashes = total_chip_hashes - previous_chip_hashes;
```
luke-jr: Tab issue.

driver-gridseed.c
```diff
@@ -239,32 +288,64 @@ void gridseed_submit_nonce(struct thr_info * const thr, const unsigned char buf[
...
 int64_t gridseed_scanhash(struct thr_info *thr, struct work *work, int64_t __maybe_unused max_nonce)
 {
+	struct timeval start_tv, nonce_range_tv, report_hashes_tv;
+	timer_set_now(&start_tv);
+
 	struct cgpu_info *device = thr->cgpu;
 	unsigned char buf[GC3355_READ_SIZE];
-	int read = 0;
 	int fd = device->device_fd;
+	int read = 0;
```
luke-jr: Why is read moved down?

driver-gridseed.c
```diff
@@ -239,32 +288,64 @@ void gridseed_submit_nonce(struct thr_info * const thr, const unsigned char buf[
...
+	int64_t total_hashes = 0;
+	bool full_range_scanned = false;
 
-	while (!thr->work_restart && (read = gc3355_read(fd, (char *)buf, GC3355_READ_SIZE)) > 0)
+	// amount of time it takes this device to scan a nonce range:
+	uint32_t nonce_range_sec = gridseed_nonce_range_duration(device);
+	// timer to break out of scanning should we scan an entire nonce range
+	timer_set_delay_from_now(&nonce_range_tv, nonce_range_sec * 1000000);
```
luke-jr: Probably should break out earlier to minimise downtime?

driver-gridseed.c
```diff
@@ -239,32 +288,64 @@ void gridseed_submit_nonce(struct thr_info * const thr, const unsigned char buf[
...
+	
+	// timer to calculate hashes every 10s
+	const uint32_t report_delay = 10 * 1000000;
+	timer_set_delay_from_now(&report_hashes_tv, report_delay);
+
+	while (!thr->work_restart                                                   // true when new work is available (miner.c)
+	    && ((read = gc3355_read(fd, (char *)buf, GC3355_READ_SIZE)) >= 0)       // only check for failure - allow 0 bytes
+	    && !(full_range_scanned = timer_passed(&nonce_range_tv, NULL)))        // true when we've had time to scan a range
```
luke-jr: Comment doesn't line up

driver-gridseed.c
```diff
@@ -239,32 +288,64 @@ void gridseed_submit_nonce(struct thr_info * const thr, const unsigned char buf[
...
 			gridseed_submit_nonce(thr, buf, work);
 		else
-		{
 			applog(LOG_ERR, "%"PRIpreprv": Unrecognized response", device->proc_repr);
-			break;
-		}
 	}
 
+	// this should never happen in practice - we should always get new work before the
+	// device even comes close to scanning a full range of nonces
```
luke-jr: Should also be exiting from scanhash so you get a new job going before it runs out... even if it does near the end.

nwoolls: Not sure what you mean here - full_range_scanned is being used to exit from scanhash. The only thing done is logging errors and reporting hashes.

Should those not be done?

luke-jr: Just saying, the loop should be exiting _before_ full_range_scanned can become true, otherwise it will be a problem should a nonce range complete.

PR #529, opened 2014-07-26
Include broad udev rules when configured with --enable-broad-udevrules option

luke-jr:


README
```diff
@@ -136,6 +136,9 @@ Optional Dependencies:
 BFGMiner driver configuration options:
 	--disable-other-drivers Build without drivers by default unless explicitly
 	                        enabled
+	--enable-broad-udevrules
+	                        Include udev rules that for ambiguous devices which
```
nwoolls: "that for" -> "for"

configure.ac
```diff
@@ -242,6 +242,17 @@ AC_ARG_ENABLE([other-drivers],
 )
 
 
+broad_udevrules=false
+AC_ARG_ENABLE([broad-udevrules],
+	[AC_HELP_STRING([--enable-broad-udevrules],[Include udev rules that for ambiguous devices which may not be miners])],
```
nwoolls: "that for"

PR #540, opened 2014-08-15
Initial version of coinbase checking funciton for stratum pool

4tar:
In stratum pool mining mode, the pool sends mining_notify message to distribute job, however, to reduce bandwidth usage, there is no transaction list included in the message, the miner would have to ask for it explicitly by sending a get_transactions message if desired, that in effect loses the bandwidth saving of stratum protocol, which actually is the essence.

By adding a coinbase checking function, we can possibly obsolete the necessary of get_transactions message during most circumstance:
    1. We ensure this is a valid BTC coinbase for btc, so reduce the possibility of pool stealing power for other kind of work;
    2. We check amount and to-address by going through the whole payout list, so reduce the possibility of pool stealing the coins mined;
    3. We can even specify the percent regards how much pool can gain from this mining job, so reduce the possibility of pool transferring (part of) coins mined to unknown address;
    4. This also creates possibility for other kind of collaborative mode between pool and miners, which we are actually looking on...

The coinbase checking can not totally replace get_transactions message, but combined with other human actions it can give miners‘ enough confidence to trust a pool.

The usage is pretty straightforward: input --coinbase-perc argument with --coinbase-addr, e.g.: --coinbase-addr='XXXX' --coinbase-perc='=1', or --coinbase-perc='>0.99', or --coinbase-perc='<0.05', etc, the command will check the addr 'XXXX' get corresponding percent of the whole benefit, else it would output an error message.

Signed-off-by: Huang Le 4tarhl@gmail.com


miner.c
```diff
@@ -1196,9 +1198,24 @@ char *set_b58addr(const char * const arg, bytes_t * const b)
...
 }
+
+static
+char *set_payperc(const char * const arg, compare_op_t * op)
+{
+	if (!bytes_len(&opt_coinbase_script))
+		return "Specify coinbase-addr before coinbase-perc";
+
+	if (arg[0] != '>' && arg[0] != '<' && arg[0] != '=')
+		return "Invalid coinbase-perc, need a '>', '<' or '=' prefix";
```
luke-jr: Is there any reason someone would ever choose something other than '>' (or rather, '>=') here?

miner.c
```diff
@@ -9353,14 +9376,14 @@ void gen_stratum_work2(struct work *work, struct stratum_work *swork)
 	data32 = (uint32_t *)merkle_sha;
 	swap32 = (uint32_t *)merkle_root;
 	flip32(swap32, data32);
-	
+
```
luke-jr, line 9465: Please don't de-indent unnecessarily...

util.c
```diff
@@ -2395,27 +2506,93 @@ static bool parse_notify(struct pool *pool, json_t *val)
...
+			pass_cb_check = (cb_perc > opt_coinbase_perc_op.value);
+		else if (opt_coinbase_perc_op.op == '<')
+			pass_cb_check = (cb_perc < opt_coinbase_perc_op.value);
+		else
+			pass_cb_check = (cb_perc == opt_coinbase_perc_op.value);
+		if (!pass_cb_check) {
+			cg_wunlock(&pool->data_lock);
+			applog(LOG_ERR, "Coinbase check: lopsided target/total = %g(%ld/%ld), expecting %c %g",
+				cb_perc, target, total, opt_coinbase_perc_op.op, opt_coinbase_perc_op.value);
+			goto out;
```
luke-jr: It's too late to goto out here, things are already modified on the stratum_work; a lock is held!

util.c
```diff
@@ -2395,27 +2506,93 @@ static bool parse_notify(struct pool *pool, json_t *val)
...
-
 	bytes_resize(&pool->swork.coinbase, pool->swork.nonce2_offset + pool->swork.n2size + cb2_len);
 	uint8_t *coinbase = bytes_buf(&pool->swork.coinbase);
 	hex2bin(coinbase, coinbase1, cb1_len);
 	hex2bin(&coinbase[cb1_len], pool->swork.nonce1, pool->n1_len);
 	// NOTE: gap for nonce2, filled at work generation time
 	hex2bin(&coinbase[pool->swork.nonce2_offset + pool->swork.n2size], coinbase2, cb2_len);
-	
+
+	if (opt_coinbase_perc_op.op) {
```
luke-jr: This is the wrong place to be doing checks - it should be somewhere generic so it checks GBT (and any future protocols) too :)

luke-jr:
Thank you for this. I've commented a few things that should probably be addressed before merging.

For review, you can keep these here, but for merging, low-level bitcoin stuff (b58enc, script_to_address, varint_decode, etc) should be in libblkmaker.

Note that while these checks are helpful and important, they don't negate the need to download transaction data, to check that - which is much more important than merely the generation transaction :) (this is not a problem with your pull request, just a comment)

miner.c
4tar: Yes I also considered to support >= kinda things but then gave up, since the comparison are between float type value so the ">=" or "<=" could almost be replaced by a more precise ">" or "<", e.g. ">=0.99" could be replaced by ">0.989" or ">0.9899".  We can save some boring code by not implement that...

miner.c
4tar, line 9465: Got it.  One of my scripts would automatically delete tab or blank chars in a 'blank' line, looks it worked when I did't notice it:)

util.c
4tar: Good point.  Let me think where to put it...  Any suggestion?

util.c
4tar: True.  A separate coinbase checking function is needed, just like what we have discussed above...

4tar:
@luke-jr Thanks for so quick response, now I'm thinking about to implement a separate coinbase checking function based on your comment.  Thanks!

Regards the low-level stuff(b58enc, script_to_addr, varint_decode, etc), yes, I agree with you that it sould be implemented in some supportive lib like libblkmaker but I just put it here for a easy review.  Actually I also think about submit patch to libblkmaker but since it named as "block maker" and only b58dec/addr_to_script/varint_encode are really needed in a block making process so I was wondering whether the counterparts are welcome:)

util.c
luke-jr: There isn't really any central place to do this check yet. Perhaps add a function and call it from both here and the GBT equivalent (inside miner.c:work_decode), before unlocking? Probably should disable the pool (add a POOL_MISBEHAVING or something to enum pool_enable) when this happens, rather than reject it outright and risk working on an old job.

miner.c
luke-jr: Right, my point is that it should be sufficient to _only_ do a >= check, and just accept a number here.

4tar:
One more thing: Yes we absolutely agree that coinbase checking could not replace txn list retrieving, actually the implemention of coinbase check is part of our work to restore the pool-miner trust, but not to negate the txn list downloading at the first place.

Thanks!

luke-jr:
Yes, I see how they technically don't really fit in libblkmaker. Perhaps we ought to split out from libblkmaker a libbase58, libbcaddress, and libbcvarint at the same time? Probably easier to just throw it in BFGMiner for the short term, I suppose.

luke-jr:
BTW, to be clear: Are you okay with those utility functions (base58, varint, script_to_addr) under MIT license terms (eg, libblkmaker)? Or just GPLv3+?

4tar:
@luke-jr I'm fine with any license you prefer:)

4tar:
@luke-jr I'm working on creating a separate function to do the coinbase checking, it almost straight forward except one thing:  currently in the lock holding area, the code will try to update the n1len and n2size, but according to the protocol, these two value should be set at the first mining_subscribe message and would not be changed during the whole session.  If this is the case, we can easily populate the coinbase block outside the lock holding area and do the checking, else the checking would have get into the locked area.

I think it's safe to code with that assumption, if you are ok with it, I'm going to do all related code change based on that.

Any thought?

luke-jr:
Nicehash has a stratum extension that changes the extranonce parameters.

4tar:
K, thanks for the information.  Will think it over again.

miner.c
4tar: Oh, actually we would have a requirement on the '<' check, so here it is:)

4tar:
@luke-jr I've separated the function into an independent function check_coinbase() and call it at the very begin of the locked area in parse_notify() when no real change had been made, it should be find now.

I also went through miner.c::work_decode() as you suggested, but looks like in which the coinbase is mostly re-built by blkmake call so I doubt whether we really need to do the check there?

util.c
```diff
@@ -2379,12 +2571,27 @@ static bool parse_notify(struct pool *pool, json_t *val)
...
+	n1_len = pool->next_nonce1 ? strlen(pool->next_nonce1) / 2 : pool->n1_len;
+	bytes_resize(&new_coinbase, cb1_len + n1_len + pool->next_n2size + cb2_len);
+	uint8_t *coinbase = bytes_buf(&new_coinbase);
+	hex2bin(coinbase, coinbase1, cb1_len);
+	hex2bin(coinbase + cb1_len + n1_len + pool->next_n2size, coinbase2, cb2_len);
+	if (!check_coinbase(coinbase, bytes_len(&new_coinbase), &opt_coinbase_perc_op, &opt_coinbase_script)) {
+		/* We can safely quit here, since no change had been made till now */
+		cg_wunlock(&pool->data_lock);
+		bytes_free(&new_coinbase);
+		goto out;
```
luke-jr: This will probably trigger a mere "Unknown stratum msg" error, and continue hashing on the old job. We could disconnect, but it's unlikely to get a "passing" job when we reconnect. I think we should just disable the pool and retest it on the next job...

luke-jr:
libblkmaker only builds a new coinbase if one is missing.

util.c
4tar: Woo... I didn't expect the coinbase checking to be such an important rule that it can decide to disable a pool... But I'm ok with it:)  So do we need to check there are other pool enabled (enabled_pools) before disabling it or just do it unconditionally?

luke-jr: Well, what other behaviour would you expect from such a violation? :) Hashing on the old job is almost certainly a bad idea, IMO. I'd just do it unconditionally - worst case, the user should failover to solo mining (which I'm assuming is active in this kind of situation).

4tar: What I thought was just giving a notice to the user and let them do whatever they decide to do... But I'm very happy to disable such a pool!!!

4tar: One more thing:  I just look into the pool disabling code and found that the global value enabled_pools is operated without any lock protection, there could be race issues, so would you like to add a lock for it or leverage another existing one?

luke-jr: Probably okay to abuse lp_lock for this since it's already held in enable_pool.

miner.c
```diff
@@ -4408,20 +4429,17 @@ static void enable_pool(struct pool *pool)
...
-#endif
-
-static void reject_pool(struct pool *pool)
+void disable_pool(struct pool *pool, enum pool_enable enable_status)
 {
-	if (pool->enabled == POOL_ENABLED)
-		enabled_pools--;
-	pool->enabled = POOL_REJECTING;
+	if (pool->enabled != POOL_ENABLED) {
+		pool->enabled = enable_status;
```
luke-jr, line 4520: Rejecting and misbehaving shouldn't override a manual/user disable here.

miner.c
```diff
@@ -4408,20 +4429,17 @@ static void enable_pool(struct pool *pool)
...
+		return;
+	}
+	mutex_lock(&lp_lock);
+	enabled_pools--;
+	pool->enabled = enable_status;
+	pthread_cond_broadcast(&lp_cond);
```
luke-jr: I'm not sure this needs a broadcast?

util.c
```diff
@@ -2379,12 +2572,32 @@ static bool parse_notify(struct pool *pool, json_t *val)
...
+	uint8_t *coinbase = bytes_buf(&new_coinbase);
+	hex2bin(coinbase, coinbase1, cb1_len);
+	hex2bin(coinbase + cb1_len + n1_len + pool->next_n2size, coinbase2, cb2_len);
+	if (!check_coinbase(coinbase, bytes_len(&new_coinbase), &opt_coinbase_perc_op, &opt_coinbase_script)) {
+		/* We can safely quit here, since no change had been made till now */
+		cg_wunlock(&pool->data_lock);
+		bytes_free(&new_coinbase);
+
+		applog(LOG_ERR, "Disable pool %d for failing to pass coinbase check", pool->pool_no);
+		disable_pool(pool, POOL_DISABLED);
```
luke-jr: POOL_DISABLED is for manual/user requests only, hence [my suggestion for a POOL_MISBEHAVING](https://github.com/luke-jr/bfgminer/pull/540#discussion_r16283934)

util.c
```diff
@@ -2351,14 +2352,206 @@ void stratum_probe_transparency(struct pool *pool)
...
+		if (varint_decode(coinbase + pos, &curr_pk_script_len) != 1 || curr_pk_script_len > 25) {
+			applog(LOG_ERR, "Coinbase check: output script too long: %ld", curr_pk_script_len);
+			return false;
+		}
+		pos += 1 /* varint length */;
+
+		if (script_to_address(addr, sizeof(addr), coinbase + pos, curr_pk_script_len, opt_testnet_addr) > sizeof(addr)) {
+			char script[51];
+			bin2hex(script, coinbase + pos, curr_pk_script_len);
+			applog(LOG_ERR, "Coinbase check: output script to an invalid address: %s", script);
```
luke-jr: scriptPubKeys don't need to be addresses, and shouldn't be expected/required to fit a specific template in general.

util.c
```diff
@@ -2351,14 +2352,206 @@ void stratum_probe_transparency(struct pool *pool)
...
+		applog(LOG_ERR, "Coinbase check: input script sig too long: %ld", len);
+		return false;
+	}
+	pos += 1 /* varint length */ + len + 4 /* 0xffffffff */; 
+
+	total = target = 0;
+
+	for (pos += varint_decode(coinbase + pos, &len); len > 0; --len) {
+		for (amount = 0, i = 7; i >= 0; --i)
+			amount = (amount << 8) + coinbase[pos + i];
```
luke-jr: upk_u64be should work here... although I thought the amount was encoded in LE?

util.c
```diff
@@ -2351,14 +2352,206 @@ void stratum_probe_transparency(struct pool *pool)
...
+		total += amount;
+		pos += 8; /* amount length */
+
+		if (varint_decode(coinbase + pos, &curr_pk_script_len) != 1 || curr_pk_script_len > 25) {
+			applog(LOG_ERR, "Coinbase check: output script too long: %ld", curr_pk_script_len);
```
luke-jr: This should not be cause of a failure.

miner.c
luke-jr: What circumstance would you want to check '<'? Furthermore, note that '<' and '=' cases are broken per this (overall) implementation: addresses are not _necessarily_ encoded the standard way (there is also the older pay-to-pubkey form).

miner.c
4tar, line 4520: Oh yes, will change in following commit.

miner.c
4tar: I couldn't be sure either, so just send it for your comment.  So we can safely remove the broadcast?

util.c
4tar: K, going to add a new status now...

util.c
4tar: Yeah it's LE, so did the code decode it as LE (i from 7 .. 0).  Will change to upk_u64le, thanks for pointing this.

api.c
```diff
@@ -2597,7 +2597,7 @@ static void removepool(struct io_data *io_data, __maybe_unused SOCKETTYPE c, cha
 		return;
 	}
 
-	pool->enabled = POOL_DISABLED;
+	disable_pool(pool, POOL_DISABLED);
```
luke-jr: Actually, let's move this into remove_pool so the same bug isn't possibly repeated later. :)

luke-jr: Note to remove the disable_pool from miner.c:display_pools when doing that.

miner.c
```diff
@@ -1196,9 +1196,24 @@ char *set_b58addr(const char * const arg, bytes_t * const b)
 		free(script);
 		return "Failed to convert address to script";
 	}
+	opt_testnet_addr = (arg[0] != '1');
```
luke-jr: This will break for Bitcoin (mainnet) addresses beginning with '3' or 'x'.

miner.c
```diff
@@ -2004,10 +2019,16 @@ static struct opt_table opt_config_table[] = {
...
+		     "Set coinbase payout address for solo mining, or stratum pool mining coinbase checking"),
 	OPT_WITH_ARG("--coinbase-address|--coinbase-payout|--cbaddress|--cbaddr|--cb-address|--cb-addr|--payout",
 		     set_b58addr, NULL, &opt_coinbase_script,
 		     opt_hidden),
+	OPT_WITH_ARG("--coinbase-perc",
+		     set_payperc, NULL, &opt_coinbase_perc_op,
+		     "Specify the percent regards how much benefit the payout address gains"),
+	OPT_WITH_ARG("--coinbase-percent|--payout-percent|--payperc",
+		     set_payperc, NULL, &opt_coinbase_perc_op,
+		     opt_hidden),
```
luke-jr: Rather than global options, I wonder if these should perhaps instead be per-pool? (see util.c:uri_find_param)

miner.c
```diff
@@ -3055,6 +3076,14 @@ static bool work_decode(struct pool *pool, struct work *work, json_t *val)
 		return false;
 	}
 	else
+	if (!check_coinbase(work->data, sizeof(work->data), &opt_coinbase_perc_op, &opt_coinbase_script)) {
```
luke-jr: work->data is a block header, not a coinbase! You want to do this down under where pool->nonce2 is reset.

miner.c
```diff
@@ -3055,6 +3076,14 @@ static bool work_decode(struct pool *pool, struct work *work, json_t *val)
...
+		applog(LOG_ERR, "Disable pool %d for failing to pass coinbase check", pool->pool_no);
+		disable_pool(pool, POOL_MISBEHAVING);
+		if (pool == current_pool())
+			switch_pools(NULL);
```
luke-jr: This code is generic regardless of protocol, so we can move it to a function too.

util.c
```diff
@@ -2351,14 +2352,224 @@ void stratum_probe_transparency(struct pool *pool)
...
+/* NOTE: The caller should ensure an valid target_script pointer if providing an valid compare_op pointer */
+bool check_coinbase(const uint8_t *coinbase, size_t cbsize, compare_op_t *compare_op, bytes_t *target_script)
+{
+	int i;
+	size_t pos;
+	uint64_t len, total, target, amount, curr_pk_script_len;
+	bool do_payout_check;
+	char addr[35];
+
+	if (cbsize < 90) { /* Smallest possible length */
```
luke-jr: I calculate 62... 4 byte version + 1 byte txin count, + 36 byte txin0 prevout + 1 byte txin0 scriptSig size + 2 byte txin0 scriptSig + 4 byte txin seq + 1 byte txout count + 8 byte txout0 value + 1 byte txout0 scriptPubKey size + 4 byte locktime

util.c
```diff
@@ -2351,14 +2352,224 @@ void stratum_probe_transparency(struct pool *pool)
...
+		applog(LOG_ERR, "Coinbase check: invalid length -- %zu", cbsize);
+		return false;
+	}
+	pos = 4; /* Skip the version */
+
+	if (varint_decode(coinbase + pos, &len) != 1 || len != 1) {
```
luke-jr: Can short-circuit this easily and safely: if (coinbase[pos] != 1)

util.c
```diff
@@ -2351,14 +2352,224 @@ void stratum_probe_transparency(struct pool *pool)
...
+		applog(LOG_ERR, "Coinbase check: multiple inputs (%ld) in coinbase", len);
+		return false;
+	}
+	pos += 1 /* varint length */ + 32 /* prevhash */ + 4 /* 0xffffffff */;
+
+	if (varint_decode(coinbase + pos, &len) != 1 || len > 100) {
```
luke-jr: if (coinbase[pos] > 100) also works

luke-jr: Or better, if (coinbase[pos] > 100 || coinbase[pos] < 2)

util.c
```diff
@@ -2351,14 +2352,224 @@ void stratum_probe_transparency(struct pool *pool)
...
+	do_payout_check = (compare_op && compare_op->op);
+
+	if (cbsize <= pos) {
+		if (do_payout_check) {
+incomplete_cb:
+			applog(LOG_ERR, "Coinbase check: incomplete coinbase for payout check");
+			return false;
+		}
+		/* Blindly believe this is a good but truncated cb */
+		return true;
```
luke-jr: How could the cb ever be truncated here safely?

util.c
```diff
@@ -2351,14 +2352,224 @@ void stratum_probe_transparency(struct pool *pool)
...
+	}
+
+	total = target = 0;
+
+	for (pos += varint_decode(coinbase + pos, &len); len > 0; --len) {
+		if (cbsize <= pos + 8 && do_payout_check)
+			goto incomplete_cb;
+
+		amount = upk_u64le(coinbase, pos);
```
luke-jr, line 2527: Out-of-range access possible if !do_payout_check

util.c
```diff
@@ -2351,14 +2352,224 @@ void stratum_probe_transparency(struct pool *pool)
...
+		pos += 8; /* amount length */
+
+		total += amount;
+
+		pos += varint_decode(coinbase + pos, &curr_pk_script_len);
```
luke-jr: Possible out-of-range access

util.c
```diff
@@ -2351,14 +2352,224 @@ void stratum_probe_transparency(struct pool *pool)
...
+		if (cbsize <= pos + curr_pk_script_len && do_payout_check)
+			goto incomplete_cb;
+
+		i = script_to_address(addr, sizeof(addr), coinbase + pos, curr_pk_script_len, opt_testnet_addr);
```
luke-jr: Out-of-range access possible if !do_payout_check

util.c
```diff
@@ -2351,14 +2352,224 @@ void stratum_probe_transparency(struct pool *pool)
...
+				if (i)
+					target += amount;
+			} else
+				i = 0;
+			if (opt_debug)
+				applog(LOG_DEBUG, "Coinbase output: %10ld -- %34s%c", amount, addr, i ? '*' : '\0');
+		} else if (opt_debug) {
+			i = (sizeof(addr) - 1) / 2;
+			if (i > curr_pk_script_len)
+				i = curr_pk_script_len;
```
luke-jr: Easier to just throw the hex on the stack: char hex[(curr_pk_script_len \* 2) + 1];

util.c
```diff
@@ -2351,14 +2352,224 @@ void stratum_probe_transparency(struct pool *pool)
...
+			bin2hex(addr, coinbase + pos, i);
+			applog(LOG_DEBUG, "Coinbase output: %10ld PK %34s%c", amount, addr, (i < curr_pk_script_len) ? '-' : '\0');
+		}
+
+		pos += curr_pk_script_len;
+	}
+	if (!total) {
+		applog(LOG_ERR, "Coinbase check: output no coins");
+		return false;
```
luke-jr, line 2565: This is valid. Perhaps instead we ought to check that the total is some expected value? Although that breaks altcoins...

util.c
```diff
@@ -2351,14 +2352,224 @@ void stratum_probe_transparency(struct pool *pool)
...
 static bool parse_notify(struct pool *pool, json_t *val)
 {
 	const char *prev_hash, *coinbase1, *coinbase2, *bbversion, *nbit, *ntime;
 	char *job_id;
 	bool clean, ret = false;
 	int merkles, i;
-	size_t cb1_len, cb2_len;
+	size_t cb1_len, cb2_len, n1_len;
 	json_t *arr;
+	bytes_t new_coinbase;
```
luke-jr: needs = BYTES_INIT;

util.c
```diff
@@ -2379,12 +2590,32 @@ static bool parse_notify(struct pool *pool, json_t *val)
...
+	if (!check_coinbase(coinbase, bytes_len(&new_coinbase), &opt_coinbase_perc_op, &opt_coinbase_script)) {
+		/* We can safely quit here, since no change had been made till now */
+		cg_wunlock(&pool->data_lock);
+		bytes_free(&new_coinbase);
+
+		applog(LOG_ERR, "Disable pool %d for failing to pass coinbase check", pool->pool_no);
+		disable_pool(pool, POOL_MISBEHAVING);
+		if (pool == current_pool())
+			switch_pools(NULL);
+		goto out;
```
luke-jr: Might as well let it finish?

4tar: To avoid the "unknown stratum message"?  Sure...

util.c
4tar: You are right, I mis-calculated mainly for always need a to_address script...

miner.c
4tar: What kind of address would have a prefix 'x'?  I'm aware of just '1' or '3'...

util.c
4tar: Not sure what kind of situation the caller would be in, so just in case...  You think we should return false for a crude truncating on cb?

miner.c
luke-jr: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#Serialization_format

util.c
luke-jr: Probably a good idea.

util.c
4tar: Just didn't want to malloc() on the heap so reuse the addr[], but I'm fine to do that if showing the whole script content in debug mode is preferred...

luke-jr: Don't need to malloc(), it's perfectly fine to put a dynamic array on the stack:

```
char hex[(curr_pk_script_len * 2) + 1];
```

luke-jr: Actually, maybe better not - Mac has a small stack, and coinbases can be large. :(

4tar: Oh my, I don't realize that C99 has implement dynamic array!  What a devil!

api.c
4tar: Please clarify which is is what you want:
1. No independent disable_pool(), any time when the user choose to disable a pool, we just do remove_pool();
2. Still going to have two separate func disable_pool() and remove_pool(), but just combine them when they appear together in the code;

luke-jr: remove_pool() is currently broken if you call it on an enabled pool. Therefore, remove_pool() should in fact call disable_pool() itself.

4tar: See.  Then I'm going to 2, right?

[Following] K, I'm just going to 2 and make a submit soon for review.

curtds: I am totally new at this and I have no idea what you are talking about or
what to do.

On Sat, Aug 16, 2014 at 8:26 PM, Huang Le notifications@github.com wrote:

> In api.c:
> 
> > @@ -2597,7 +2597,7 @@ static void removepool(struct io_data *io_data, __maybe_unused SOCKETTYPE c, cha
> >         return;
> >     }
> > -   pool->enabled = POOL_DISABLED;
> > -   disable_pool(pool, POOL_DISABLED);
> 
> Please clarify which is is what you want:
> 1. No independent disable_pool(), any time when the user choose to
>    disable a pool, we just do remove_pool();
> 2. Still going to have two separate func disable_pool() and
>    remove_pool(), but just combine them when they appear together in the code;
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/luke-jr/bfgminer/pull/540/files#r16328016.

## 

"I can do all things through Christ which strengtheneth me" (Phil 4:13).

Jerry and God http://www.jerryandgod.com/

Biblical People http://www.biblicalpeople.wordpress.com/

Jerry and God: The Mighty Hunters
http://www.slideboom.com/presentations/698053/Jerry-and-God%3A-The-Mighty-Hunters%2C-Book-1?pk=c0ec-6052-c01e-c608-d056-b3aa-f4f4-dd40

Virtual Assistant http://curtstowellvirtualassistant.wordpress.com/

luke-jr: No problem, I'll do it - just thought it was a convenient time to :)

luke-jr: Err, @curtds isn't @4tar ... :)

@4tar Right, basically just stick a call to disable_pool at the top of remove_pool and adjust remove_pool uses.

@curtds I'm guessing you subscribed to the project, but wanted to subscribe to releases? For the latter, use https://lists.nongnu.org/mailman/listinfo/bfgminer-announce

util.c
```diff
@@ -2252,6 +2253,20 @@ const char *uri_find_param(const char * const uri, const char * const param, boo
 	return URI_FIND_PARAM_FOUND;
 }
 
+size_t uri_get_param(const char * const uri, const char * const param, char *val, size_t size)
+{
+	const char *p, *q = uri_find_param(uri, param, NULL);
+	if (q && q != URI_FIND_PARAM_FOUND) {
+		for (p = q; *p && *p != ','; ++p);
```
luke-jr: Delimiter might be other than a comma, like another '#'

util.c
```diff
@@ -2351,14 +2366,254 @@ void stratum_probe_transparency(struct pool *pool)
...
+{
+	int i;
+	size_t pos;
+	uint64_t len, total, target, amount, curr_pk_script_len;
+	bool on_testnet = false, found_target = false;
+	bytes_t target_script = BYTES_INIT;
+	char addr[35];
+
+	if (cbsize < 62) { /* Smallest possible length */
+		applog(LOG_ERR, "Coinbase check: invalid length -- %zu", cbsize);
```
luke-jr: %zu is not valid on all supported platforms, please cast to unsigned long.

util.c
```diff
@@ -2351,14 +2366,254 @@ void stratum_probe_transparency(struct pool *pool)
...
+		applog(LOG_ERR, "Coinbase check: incomplete coinbase for payout check");
+		return false;
+	}
+
+	if (target_addr) {
+		if (set_b58addr(target_addr, &target_script)) {
+			applog(LOG_ERR, "Coinbase check: against an invalid addr: %s", target_addr);
+			return false;
+		}
+		on_testnet = target_addr[0] != '1' && target_addr[0] != '3' && target_addr[0] != 'x';
```
luke-jr: 'x' could be testnet too :(

But I guess this doesn't matter (yet) since the rest of the code here doesn't support 'x' addresses. :)

util.c
```diff
@@ -2351,14 +2366,254 @@ void stratum_probe_transparency(struct pool *pool)
...
+	if (cbsize <= pos) {
+incomplete_cb:
+		applog(LOG_ERR, "Coinbase check: incomplete coinbase for payout check");
+		return false;
+	}
+
+	if (target_addr) {
+		if (set_b58addr(target_addr, &target_script)) {
+			applog(LOG_ERR, "Coinbase check: against an invalid addr: %s", target_addr);
+			return false;
```
luke-jr: Might be better to make this a warning and effectively disable the check. After all, it's not the pool's fault if the user puts something invalid...

util.c
```diff
@@ -2351,14 +2366,254 @@ void stratum_probe_transparency(struct pool *pool)
...
+				i = 0;
+			if (opt_debug)
+				applog(LOG_DEBUG, "Coinbase output: %10ld -- %34s%c", amount, addr, i ? '*' : '\0');
+		} else if (opt_debug) {
+			char *hex = addr;
+			if (curr_pk_script_len * 2 >= sizeof(addr))
+				hex = malloc(curr_pk_script_len * 2 + 1);
+			if (hex) {
+				bin2hex(hex, coinbase + pos, curr_pk_script_len);
+				applog(LOG_DEBUG, "Coinbase output: %10ld PK %34s", amount, hex);
```
luke-jr: What's the purpose of the hex padding?

util.c
```diff
@@ -2351,14 +2366,254 @@ void stratum_probe_transparency(struct pool *pool)
...
+		i = varint_decode(coinbase + pos, cbsize - pos, &curr_pk_script_len);
+		if (!i || cbsize <= pos + i + curr_pk_script_len)
+			goto incomplete_cb;
+		pos += i;
+
+		i = script_to_address(addr, sizeof(addr), coinbase + pos, curr_pk_script_len, on_testnet);
+		if (i && i <= sizeof(addr)) { /* So this script is to payout to an valid address */
+			if (target_addr) {
+				i = (bytes_len(&target_script) == curr_pk_script_len &&
+					!memcmp(bytes_buf(&target_script), coinbase + pos, curr_pk_script_len));
```
luke-jr: This won't match the old pay-to-pubkey script form. Is that intentional? It seems to contradict the '<' and '=' comparisons.

util.c
```diff
@@ -2351,14 +2366,254 @@ void stratum_probe_transparency(struct pool *pool)
...
+
+static inline bool do_compare(double value, compare_op_t *op)
+{
+	switch (op->op) {
+	case '>':
+		return (value > op->value);
+	case '<':
+		return (value < op->value);
+	case '=':
+		return (value == op->value);
```
luke-jr: What is the use case for the '<' and '=' comparisons?

luke-jr:
FWIW, merged parts of this in d338c29..3bd3f9b

util.c
4tar: No special meaning, just try to align all debug output, would it be confusing?

util.c
4tar: Wow, then I'd just leave it there and make a comment, maybe add support for 'x' addr later...

util.c
4tar: Besides the flexibility for users, we do have cases for '=' and '<'. '=' is usual, for the pool receives all rewards in the coinbase before it make payment to miners later; '<' is for a special case we are thinking, in which several miners who are running quite huge a farm and connect into a pool hosted for this farm, they wanna be sure that the pool receives say less than 1 percent of rewards.

The latter case would be better addressed by letting the check supports a list of address and still uses '>' or '=' to do the comparison, but that would complicate things so '<' comes into as a provisional/crude alternative...

util.c
4tar: That's true!  Going to change as you say.

util.c
4tar: Hmm, is the pay-to-pubkey still used?  I'm going to check the OP_CHECK_SIG payout script as well.  Any other possibility missed by the code?

luke-jr: It's not still used, but right now it'd be a convenient way to bypass the '<' and '=' comparisons.

util.c
luke-jr: What stops the pool from simply using a bunch of other addresses in addition to this one? Or even simply adding an OP_NOP on the end?

4tar: Sure, that's why the real requirement is to support a list of address and comparing against the sum of rewards all of those address in that list are receiving.  Hmm, maybe we should add such support now instead of introducing this '<' things...

luke-jr: I think that would be better... :)

util.c
```diff
@@ -2351,14 +2377,329 @@ void stratum_probe_transparency(struct pool *pool)
...
+	const char *target_addr, compare_op_t *cbtotal_compare_op, compare_op_t *cbperc_compare_op )
+{
+	int i, addr_c = 0;
+	size_t pos;
+	uint64_t len, total, target, amount, curr_pk_script_len;
+	bool on_testnet = false, found_target = false, ret = false;
+	char **addr_v = NULL;
+
+	if (cbsize < 62) { /* Smallest possible length */
+		applog(LOG_ERR, "Coinbase check: invalid length -- %lu", cbsize);
```
luke-jr, line 2491: cbsize is size_t here, not unsigned long. You need a cast...

util.c
```diff
@@ -2351,14 +2377,329 @@ void stratum_probe_transparency(struct pool *pool)
...
+
+	if (cbsize < pos + 4) {
+		applog(LOG_ERR, "Coinbase check: No room for locktime");
+		goto out;
+	}
+	pos += 4;
+
+	if (opt_debug)
+		applog(LOG_DEBUG, "Coinbase: (size, pos, addr_count, target, total) = (%lu, %lu, %d, %lu, %lu)",
+			cbsize, pos, addr_c, target, total);
```
luke-jr: More type mismatches here

luke-jr:
At this point, we're making very long pool URIs. What do you think of having a separate pool parameter for the list? This would also make it easier to just build it as a hash table for faster checking...

4tar:
@luke-jr The original ripemd160 code was C++ file I took from Bitcoin Core Project which has a MIT license, just a simply re-writing in C.  Regards any other code I wrote, please just put any license you prefer, I'm always fine.

Yes, that should be a "OR" instead of "AN" when checking script[0] be 2 or 3, my fault!  And for the value in (3,8), aka 4,5,6,7, which indicates the pubkey version for an uncompressed 65-bytes pubkey, while 2,3 for 33-bytes compressed pubkey.

Having separate parameter sounds great, I do also feel the pool URI is getting annoying long with all these things put together.  Let me follow on this idea...

luke-jr:
For pay-to-pubkey, script[0] should be a push opcode (which IIRC could potentially be a single-byte push-by-length, OP_PUSHDATA1, OP_PUSHDATA2, or OP_PUSHDATA4). It's important to verify all the opcodes are the same, so we don't recognise other non-standard scripts as pay-to-pubkey or pay-to-pubkeyhash too. Additionally, I'm not sure we can assume that the pubkey used is not padded in some way that allows it to pass verification - so we really need to check this case regardless of the push length. THAT BEING SAID, since we're moving to a >= check, it may not be a bad idea to only look for pay-to-pubkeyhash, since pay-to-pubkey is obsolete and may not be recognised as a payment by some newer wallets. So, feel free to avoid implementing all this logic by simply doing a direct script comparison again. :)

miner.c
```diff
@@ -1566,6 +1571,89 @@ static char *set_userpass(const char *arg)
...
 
+static char *set_cbcaddr(char *arg)
+{
+	struct pool *pool;
+	char *p = arg, *addr;
+	bytes_t target_script = BYTES_INIT;
+
+	total_cbaddrs++;
+	if (total_cbaddrs > total_pools)
+		add_pool();
```
luke-jr: Probably better to behave like pool-priority and simply attach to the last configured pool always, rather than requiring a cbcaddr for every pool.

4tar:
Right, now we can only check the most common case simply because we are sticking to the '>=' case.  Wow, it feels like a release from those complex and fallible logic!

miner.c
4tar: The code doesn't ask from cbaddr for every pool, since it will skip the previous pool for cbcadd/cbctotal/cbcperc arguments in add_pool().  But you did point out the better way to do this, will make the change to behave just like pool-priority arg setting.

luke-jr:
:)

BTW, if you leave a comment when you're done making changes, I can review them faster - GitHub doesn't tell me when you add commits, only when you comment :)

4tar:
@luke-jr Just push a commit to implement the three coinbase check argumenet attaching as you suggested.

And regards to the payout addr comparison, I feel no need to change current code from addr hash table lookup to script comparison or script hash table lookup, since now we don't care about the absolute correctness of other kind of non-standard payout script to address conversion, and we would need get the address from script anyway in debug mode for displaying, which always need a script_to_address() call (and again, we don't care about it might fail due to non-standard payout address).

4tar:
And I'll continue to improve the script_to_address() function but it doesn't affect our current usage of it...

miner.c
```diff
@@ -1566,6 +1563,89 @@ static char *set_userpass(const char *arg)
...
+	 */
+	pool->cb_param.testnet = (p[0] != '1' && p[0] != '3' && p[0] != 'x');
+
+	while (p) {
+		struct addr_hash *ah;
+		char *addr = p;
+
+		p = strchr(p, ',');
+		if (p)
+			*p++ = '\0';
```
luke-jr: strtok_r?

luke-jr:
If we just index scripts, we can drop all the new base58 and ripemd160 stuff I think, no? :)

miner.c
4tar: Not sure it is being on every platform... do you think we'd better to switch to it?

luke-jr: We're already using it multiple places, and have a gnulib implementation for platforms missing it.

4tar:
In debug mode we show the payout address list for user to check... If it is considered to be not need, then yes we can drop those base58/ripemd160 stuff.

miner.c
4tar: K, then I update to call the it instead.

luke-jr:
That's a good point. Why don't we just drop the p2pk (and therefore ripemd) and leave the debug for p2pkh ones?

4tar:
Sure we can.  Let me clear things up.

4tar:
Eliminate the ripemd160 related code by removing support for p2pk script.  Also switch to strtok_r() from strchr().

luke-jr:
I've moved toward integrating the base58 code in #546. It's rather invasive to the build system, so please help review.

luke-jr:
Following the merge of libbase58, I've flattened the remainder of this into d416366ca3e4016919dd45ca6494562e757b1905 - please give it a look over before merging.

luke-jr:
Superceded by $551, please review

PR #547, opened 2014-08-24
Allow Gridseed Blades and Orbs to peacefully coexist in one mining insta...

slax0r:
Allow Gridseed Blades and Orbs to peacefully coexist in one mining instance.  Also sets more useful default clock frequencies for Blade and Orb.  These frequencies are at the top of the performance range and do not trigger hardware errors on either device.  Frequencies are set dynamically per device at runtime..

### DIFF
--- a/driver-gridseed.c
+++ b/driver-gridseed.c
@@ -18,7 +18,10 @@
 #include "lowl-vcom.h"
 #include "gc3355.h"
 
-#define GRIDSEED_DEFAULT_FREQUENCY  600
+/* These are stable defaults for the ORB and BLADE (ie High speed and no HW errors)
+   seperated here to allow blades and orbs to peacefully coexist in one miner instance */
+#define GRIDSEED_ORB_DEFAULT_FREQUENCY  875
+#define GRIDSEED_BLADE_DEFAULT_FREQUENCY  825
 // 60Kh/s at 700MHz in ms
 #define GRIDSEED_HASH_SPEED         0.08571428571429
 // GridSeed driver currently scans a full nonce range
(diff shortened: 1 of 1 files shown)


luke-jr:
1) The point where hardware errors occur varies based on environment, and individual unit.
2) Even when hardware errors are not occurring, it is possible a too-high frequency may damage chips.
3) Latest release does not correctly report hardware errors on scrypt devices(!) - this is fixed in git somewhat, but still requires a high sample period to ensure accuracy.
For the above reasons, I think it's best if we stick to manufacturer-supported defaults.

Anyhow, the code already seems to support both Blades and Orbs in the same instance - your patch only changes default frequencies. If you are concerned about setting different settings based on device information (a problem which affects more than just GridSeed), I suggest extending the --set syntax in some sensible way.

slax0r:
I understand the concern there with manufactures defaults.  I had actually started off by extending the --set syntax however when orbs and blades are mixed then they don't always initialize in the same order.  I was thinking that I could apply settings based on serial number.

luke-jr:
You can already do that, if they have serial numbers... --set gsd@serialno:clock=N

slax0r:
Yep, thats what I tried at first, but it actually has no effect.

example:

sudo bfgminer --set \
GSD@48EF67453734:clock=825,\
GSD@6D97537E4853:clock=850,\
GSD@6D882F705251:clock=875

All 3 devices are initialized to the default frequency of 600.
The first device is a Gridseed Blade, the other two are Gridseed Orbs.

luke-jr:
Commas have no special meaning to --set, try without.

slax0r:
so I changed it to:

sudo bfgminer --set \
GSD@48EF67453734:clock=825\
GSD@6D97537E4853:clock=850\
GSD@6D882F705251:clock=875

Still no effect

luke-jr:
concatenating them (which is all \ does) doesn't work either... you need multiple --set options

slax0r:
So this works with multiple --sets.

sudo bfgminer \
--set gridseed@48EF67453734:clock=825 \
--set gridseed@6D97537E4853:clock=850 \
--set gridseed@6D882F705251:clock=875

Adding it to the bfgminer.conf file looks like this:

"set" : [
        "gridseed@48F165523734:clock=825",
        "gridseed@48EF67453734:clock=825",
        "gridseed@6D882F705251:clock=875",
        "gridseed@6D97537E4853:clock=850",
        "gridseed@6D8932715251:clock=875",
        "gridseed@6D7836954857:clock=875",
        "gridseed@6D85278F5650:clock=875",
        "gridseed@6D7A23A64857:clock=875",
        "gridseed@6D72186A4853:clock=875"
]

And works great. May want to add this into the README.ASIC file.  Just a suggestion.  Thanks for the guidance and thanks for making bfgminer, and for making it so flexible.

luke-jr:
Patches welcome to README.ASIC too :)

slax0r:
do it right now

PR #548, opened 2014-08-24
Update to README.ASIC to add a section for Gridseed

slax0r:
Update to README.ASIC to add a section for Gridseed including command line options/arguments and sample block for adding to bfgminer.conf for persistence.

### DIFF
--- a/README.ASIC
+++ b/README.ASIC
@@ -257,6 +257,43 @@ If you have different devices using the bifury driver, see the section on
 Hex*Fury for applying workarounds per-device.
 
 
+GRIDSEED
+--------
+
+Gridessed units at the present time come in two versions:  Blade - a 40 chip unit and Orb - a 5 chip unit.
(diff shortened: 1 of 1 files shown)


README.ASIC
```diff
@@ -257,6 +257,43 @@ If you have different devices using the bifury driver, see the section on
 Hex*Fury for applying workarounds per-device.
 
 
+GRIDSEED
+--------
+
+Gridessed units at the present time come in two versions:  Blade - a 40 chip unit and Orb - a 5 chip unit.
```
luke-jr, line 263: typo

README.ASIC
```diff
@@ -257,6 +257,43 @@ If you have different devices using the bifury driver, see the section on
...
+Orb units can be used to mine both SHA256 and Scrypt based coins whereas the Blade is Scrypt only.  BFGMiner
```
luke-jr, line 264: Probably not fair to mention the Orb can do SHA256 without also mentioning BFGMiner does not support it in this mode :)

PR #549, opened 2014-08-24
Update to README.ASIC to add a section for Gridseed

slax0r:
Update to README.ASIC to add a section for Gridseed including command line options/arguments and sample block for adding to bfgminer.conf for persistence.


luke-jr:
Merged as babbe50, thanks!

PR #555, opened 2014-09-13
Bugfix: icarus: the STATS RPC API call crashes with a multi-processor device

nwoolls:


luke-jr:
Easier to just always set info = cgpu->device->device_data;? Also, the commit message is confusing - it suggest you're introducing a crash :p

nwoolls:
@luke-jr updated this one based on your comments

PR #563, opened 2014-10-01
KnCMiner Titan driver

luke-jr:


dvitas:
Hi Luke!
I have tested your branch, it works ok on the actual hardware, no problem here, you can merge it into your master branch.
The only nitpick: I prefer to have uthash as a submodule also, and not to have manual instructions of downloading it (see https://github.com/KnCMiner/bfgminer/commit/a42ded0a2f361598c371fa44a235fc9b3caeba30)
That is not actually a big issue, you can merge this patchset as it is, without "uthash as a submodule" commit.

Thank you for your efforts! :)
Vitalii

luke-jr:
@dvitas Thanks for the review. I prefer to keep dependencies separate when possible. uthash is conveniently available as a package on most systems, and easily added in manually for the few exceptions.

PR #571, opened 2014-10-30
Add BFx2 USB miner instructions to README.ASIC

tbartlett0:
Using this miner is not obvious since it's necessary to install the WinUSB driver and then pass the -S bfx:all option. I have taken the instructions from the bitcointalk thread (https://bitcointalk.org/index.php?topic=524521.msg6194413) and added them to the README.ASIC file.

### DIFF
--- a/README.ASIC
+++ b/README.ASIC
@@ -88,6 +88,20 @@ Then you must run BFGMiner as root, with the proper driver selected.
 For example:
     sudo bfgminer -S bfsb:auto
 
+BFx2 USB
+--------
+You will need to install the WinUSB driver instead of the default FTDI serial
+driver. The easiest way to do this is using Zadig: http://zadig.akeo.ie/
+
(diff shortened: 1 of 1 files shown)


README.ASIC
```diff
@@ -88,6 +88,20 @@ Then you must run BFGMiner as root, with the proper driver selected.
 For example:
     sudo bfgminer -S bfsb:auto
 
+BFx2 USB
+--------
+You will need to install the WinUSB driver instead of the default FTDI serial
+driver. The easiest way to do this is using Zadig: http://zadig.akeo.ie/
+
+(See https://bitcointalk.org/index.php?topic=524521.msg6194413#msg6194413)
```
luke-jr, line 96: Can you take this line out? I think the rest covers it all...

luke-jr, line 96: Actually, I should fix the metadata, so I'll do this at the same time.

luke-jr:
Merged as da59f1c, thanks!

PR #582, opened 2014-11-23
Fixing column spacing of non-curses device output

Krellan:
Device and processor names of different length (ABC1 vs. ABC12)
were not having enough trailing spaces added,
so the columns were not lining up consistently for per-device stats
on non-curses output (curses output was already OK).


luke-jr:
Not sure non-curses output was ever intended to line up... I guess it's harmless to do so, but I'll leave this open for others to potentially comment on until release time.

Krellan:
Thanks, seems reasonable to me.

Josh

PR #583, opened 2014-11-23
Adding additional boolean keywords enable and disable

Krellan:
Adding yet another pair of English words that are commonly used
as Boolean directives.


Krellan:
Thanks!

Josh

PR #584, opened 2014-11-23
Shut up needless compiler warnings

Krellan:
Suppressing warnings about unused results.
Merely casting the results to void isn't good enough to keep fooling
newest compiler versions, so instead using an empty if statement.

### DIFF
--- a/driver-bifury.c
+++ b/driver-bifury.c
@@ -419,7 +419,8 @@ void bifury_handle_cmd(struct cgpu_info * const dev, const char * const cmd)
 	{
 		// job <jobid> <timestamp> <chip>
 		const uint32_t jobid = strtoll(&cmd[4], &p, 0x10);
-		strtoll(&p[1], &p, 0x10);
+		if (strtoll(&p[1], &p, 0x10))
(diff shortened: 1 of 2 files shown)


luke-jr:
This sounds like a libc bug. It's perfectly valid to not use the result from strtol...

Krellan:
Interesting.

On Linux Mint (gcc 4.8) there's no warning, but on an older Ubuntu (gcc 4.6):

util.c: In function ‘bfg_strtobool’:
util.c:2223:8: warning: ignoring return value of ‘strtol’, declared with attribute warn_unused_result [-Wunused-result]
driver-bifury.c: In function ‘bifury_handle_cmd’:
driver-bifury.c:422:10: warning: ignoring return value of ‘strtoll’, declared with attribute warn_unused_result [-Wunused-result]

Good, you're right, it was something that only affected certain distributions.  Probably not necessary to include going forward.

Thanks for taking a look at it, though!

luke-jr:
The libc version is likely to be more relevant here - it's based on whether the function is tagged a certain way.

Issue #589, opened 2014-12-09
Gridseed hash reports overflow 64bit int, avg hash rate shoots up to Ph/s


SmittyHalibut:
When start-up works (will submit a separate bug), bfgminer starts mining my two gridseeds happily.  Debug logs show ~692000 hashs every 10 seconds, 68.5khash/sec.  But, once in a while, I'll get an update that is a number just shy of 2^64.  For example:

----- snip! -----
 [2014-12-06 21:55:43] 20s:688.5 avg:684.7 u:  0.0 kh/s | A:0 R:0+0(none) HW:0/none
 [2014-12-06 21:55:50] [thread 0: 692572 hashes, 68.5 khash/sec]
 [2014-12-06 21:55:50] [thread 5: 692640 hashes, 68.5 khash/sec]
 [2014-12-06 21:55:50] [thread 1: 692572 hashes, 68.5 khash/sec]
 [2014-12-06 21:55:50] [thread 6: 692640 hashes, 68.5 khash/sec]
 [2014-12-06 21:55:50] [thread 2: 692572 hashes, 68.5 khash/sec]
 [2014-12-06 21:55:50] [thread 7: 692640 hashes, 68.5 khash/sec]
 [2014-12-06 21:55:50] [thread 3: 692572 hashes, 68.5 khash/sec]
 [2014-12-06 21:55:50] [thread 8: 692640 hashes, 68.5 khash/sec]
 [2014-12-06 21:55:50] [thread 4: 692572 hashes, 68.5 khash/sec]
 [2014-12-06 21:55:50] [thread 9: 692640 hashes, 68.5 khash/sec]
 [2014-12-06 21:56:00] [thread 5: 18446744073415732165 hashes, 1826396199456651.0 khash/sec]
 [2014-12-06 21:56:00] [thread 0: 18446744073415732234 hashes, 1826377935818338.5 khash/sec]
 [2014-12-06 21:56:00] [thread 6: 18446744073415732165 hashes, 1826340867180468.8 khash/sec]
 [2014-12-06 21:56:00] [thread 1: 18446744073415732234 hashes, 1826334900174491.0 khash/sec]
 [2014-12-06 21:56:00] [thread 7: 18446744073415732165 hashes, 1826345930125208.8 khash/sec]
 [2014-12-06 21:56:00] [thread 2: 18446744073415732234 hashes, 1826340143904940.2 khash/sec]
[...]

PR #590, opened 2014-12-09
Put units on all numbers, not one combined unit for all numbers.

SmittyHalibut:
Added units to every formatted number to allow for different units
on largely different numbers.  This prevents one out-of-whack number
from setting the unit so that the other numbers become hidden in
rounding errors.  It does take one more character per number, though.

https://github.com/luke-jr/bfgminer/issues/589

### DIFF
--- a/miner.c
+++ b/miner.c
@@ -3925,19 +3925,11 @@ char *_multi_format_unit(char **buflist, size_t *bufszlist, bool floatprec, cons
 		delimsz = strlen(delim);
 	
 	for (i = 0; i < count; ++i)
-		if (numbers[i] != 0)
-		{
-			pick_unit(numbers[i], &unit);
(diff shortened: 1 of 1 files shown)


luke-jr:
I'm not sure this is desirable? Outside of the bug scenario mentioned in #589, when would you want different units for the same hashrate measured different ways? It also takes up more screen space...

SmittyHalibut:
I'd argue that since they are measured differently, they aren't the same number at all. Granted, on average they should tend to the same value, but even without my bug, the last number can be wildly off from the others. 

It does take two additional characters on screen, that is true. But it's still less than 80 characters.

luke-jr:
The last one should never be wildly off from the others in ordinary usage...

PR #595, opened 2015-01-16
Fixes regarding wrong statistics display

dvitas:
Two fixes of wrongly displayed statistics.
1) First commit fixes wrong numbers in "detail level:processors" view (for Titans, which are the only devices using get_master_rolling_hashrate). First processor shows wrong rolling hashrate in that view;
2) Second commit fixes general problem of using incorrect difficulty when calculating diff_stale.

### DIFF
--- a/miner.c
+++ b/miner.c
@@ -395,13 +395,9 @@ static int include_count;
 #define JSON_LOAD_ERROR_LEN strlen(JSON_LOAD_ERROR)
 #define JSON_MAX_DEPTH 10
 #define JSON_MAX_DEPTH_ERR "Too many levels of JSON includes (limit 10) or a loop"
-<<<<<<<
+#define JSON_WEB_ERROR "WEB config err"
 
 char *cmd_idle, *cmd_sick, *cmd_dead;
-|||||||
-=======
-#define JSON_WEB_ERROR "WEB config err"
->>>>>>>
 
 #if defined(unix) || defined(__APPLE__)
 	static char *opt_stderr_cmd = NULL;
@@ -2908,6 +2904,7 @@ static char *load_web_config(const char *arg)
 {
 	json_t *val;
 	CURL *curl;
+	struct bfg_loaded_configfile *cfginfo;
 
 	curl = curl_easy_init();
 	if (unlikely(!curl))
@@ -2920,12 +2917,15 @@ static char *load_web_config(const char *arg)
 	if (!val || !json_is_object(val))
 		return JSON_WEB_ERROR;
 
-	if (!cnfbuf)
-		cnfbuf = strdup(arg);
+	cfginfo = malloc(sizeof(*cfginfo));
+	*cfginfo = (struct bfg_loaded_configfile){
+		.filename = strdup(arg),
+	};
+	LL_APPEND(bfg_loaded_configfiles, cfginfo);
 
 	config_loaded = true;
 
-	return parse_config(val, true);
+	return parse_config(val, true, &cfginfo->fileconf_load);
 }
 
 static char *load_config(const char *arg, void __maybe_unused *unused)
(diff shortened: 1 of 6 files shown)


miner.c
```diff
@@ -10508,8 +10505,11 @@ enum test_nonce2_result _test_nonce2(struct work *work, uint32_t nonce, bool che
 			{
 				applog(LOG_DEBUG, "Stratum pool %u target has changed since work job issued, checking that too",
 				       pool->pool_no);
-				if (hash_target_check_v(work->hash, pool->next_target))
+				if (hash_target_check_v(work->hash, pool->next_target)) {
 					high_hash = false;
+					memcpy(work->target, pool->next_target, sizeof(work->target));
+					calc_diff(work, 0);
```
luke-jr: Are you sure there are no side effects to doing this? This is an error situation that should never happen... it means the pool is broken (not following the stratum spec).

miner.c
```diff
@@ -10508,8 +10505,11 @@ enum test_nonce2_result _test_nonce2(struct work *work, uint32_t nonce, bool che
 			{
 				applog(LOG_DEBUG, "Stratum pool %u target has changed since work job issued, checking that too",
 				       pool->pool_no);
-				if (hash_target_check_v(work->hash, pool->next_target))
+				if (hash_target_check_v(work->hash, pool->next_target)) {
```
luke-jr: Prefer if the { is on a new line in the future.

miner.c
dvitas: I am not completely sure that there are no side effects, but I am pretty sure that when this happens and later those shares are discarded as stale, diff_stale is calculated absolutely incorrectly.
Also, the pool does not need to be broken. The code basically says: shares are good to be sent to pool if they comply with work->work_difficulty OR pool->next_target.
Let us consider the following scenario:
- nonce_diff is 1. pool starts with difficulty 8. That difficulty is written down as work->work_difficulty. Then pool changes difficulty to 1. pool->next_target = 1. Then at some moment a nonce is reported with diff 1.5. Test_nonce will return TNR_GOOD, because of pool->next_target.
  - pool connection breaks, miner is unable to send that share to the pool. A lot of those shares. All of them are cleared after a timeout. And they are accumulated in stale_diff counter with weight 8, although in reality they were checked against diff 1! Now we have stale_diff 8 times greater then it should be.

That is a scenario I am aiming with that change. I did not see negative side effects in my experiments, but again, I am not completely sure that there are no any.

miner.c
dvitas: Sorry for that, I forgot that you have different code style :)
Feel free to fix it if you are ok with a change itself.

miner.c
luke-jr: Ah, makes sense... let me think on this.

dvitas: Okay, more input on this matter. My recent experience with Neptune shows that this logic (retroactive application of the newly set pool difficulty to old works) brings only troubles with well-behaving pools (most of Bitcoin pools are well-behaving). It only leads to excessive rejects, because normal pools expect those old works to satisfy the old difficulty, not the newly set one.
So, at least in Bitcoin case where we have "normal" pools I found that disabling that logic gave me greatly reduced rate of rejects. So, in the next Neptune firmware release, that will be disabled by default with a help of newly introduced command-line option. What do you think of adding that option to the mainline? The code is here:
https://github.com/KnCMiner/bfgminer/commit/bac9098cfc1a70a542bf14de739dea3ef1b69c37

luke-jr: Rather do this kind of thing as a pool option, and probably makes sense to have it disabled by default. However, I thought BFGMiner was ignoring when these were rejected - although maybe as a side effect of your change here it stopped ignoring them...?

dvitas: 1) Agree with making it a pool option, will do;
2) I'd made it enabled by default only to keep the current behaviour unchanged. If you are ok with changing the default behaviour, I am ok too;
3) Maybe they are not being ignored anymore as a side effect of my change, I can't tell. But even if BFGMiner ignores them, pool does not and user still gets nasty high reject counters on the pool side. So: a) pool is broken and then there are no rejects, or b) pool is not broken and then even if BFGMiner ignores rejects, they are still being counted on pool side. I would say, there is no point in ignoring them at all.

dvitas:
And here is it, new pool option.

luke-jr:
Pool parameters should be using eg uri_get_param_bool(pool->rpc_url, "tlsca", false), not dedicated CLI params.

dvitas:
Do you really want string parsing routines to be invoked each time we are verifying a nonce? No, I don't think so...
Of course, we could parse it only once when pool parameters are updated. Can you point me to the best place where I can parse newly updated url string and remember the boolean for later use?

luke-jr:
Of course not, please see c113a75c3c485d67da918bb35b5a6027638fdd3b for how to do this in pool_set_uri (miner.c). :)

dvitas:
New pull request created with an URI option

PR #599, opened 2015-02-11
Latest kncasic fixes from kncasic-neptune branch

dvitas:


driver-kncasic.c
```diff
@@ -72,6 +72,7 @@ struct knc_core_state {
 		int slot;
 		uint32_t nonce;
 	} last_nonce;
+	uint32_t last_nonce_verified;
```
luke-jr, line 75: Is it normal for duplicate nonces?? :/

driver-kncasic.c
```diff
@@ -251,6 +250,172 @@ static int knc_transfer_completed(struct knc_state *knc, int stamp)
 	return (int)(knc->read_buffer_count - stamp) >= 1;
 }
 
+static struct cgpu_info * all_cgpus[KNC_MAX_ASICS][KNC_MAX_DIES_PER_ASIC] = {{NULL}};
```
luke-jr: This probably shouldn't be here...

driver-kncasic.c
dvitas, line 75: I am not sure what is your point here... Can you please elaborate?

luke-jr, line 75: If duplicate nonces are normal, it makes sense to detect and suppress them. If they're indicative of a problem or rare, it may be better to let the user find out by not filtering them.

dvitas, line 75: The majority of those duplicates is caused by communication errors due to external noise and crosstalks in the ribbon cable. So, basically they are expected; that's why I want to filter them out.
As to how frequent they are: in some setups I have seen up to 2% of duplicates, depending on level of external noise.
That exposure to the communication errors is something that was expected. We have checksums in the communication protocol, communication errors are being detected, but in kncasic driver we decided to process errored messages as well, just in hope that some of the reported nonces are still valid and we want to deliver valid nonces to the pool as soon as possible. So, in some sense filtering of duplicates is a tall we pay for ASAP delivery of golden nonce.

driver-kncasic.c
dvitas: Moved into struct knc_state

driver-kncasic.c
```diff
@@ -72,6 +72,7 @@ struct knc_core_state {
 		int slot;
 		uint32_t nonce;
```
dffp, line 73: Flush

luke-jr:
What's the status of this? Should I merge it?

driver-kncasic.c
```diff
@@ -155,8 +155,7 @@ struct knc_state {
 	/* lock to protect resources between different threads */
 	pthread_mutex_t state_lock;
 
-	/* Do not add anything below here!! core[] must be last */
-	struct knc_core_state core[];
+	struct knc_core_state core[KNC_MAX_ASICS * KNC_MAX_DIES_PER_ASIC * KNC_MAX_CORES_PER_DIE];
 };
```
ikeedge, line 161: be happy

ikeedge (approved):

MONIMAKER365 (approved):

Issue #608, opened 2015-03-12
autogen.sh fails cloning libblkmaker


wilson0x4d:
I have tried this on three different machines now (x86 and ARM architectures) and get the same result.

It appears the remote for 'libblkmaker' refers to gitorious using the git: protocol (rather than HTTPS) and it may be that gitorious is fialing to authorize anonymous users. If I manually clone from the command-line I can only clone if I use 'HTTPS'.

This appears to work around the issue, but I wind up having to do this every time I clone bfgminer for build. Would be nice if the origin configured in the repo could be corrected so it works for the widest audience.

Thanks. 

Here are my commands and the output:

```
root@buildsrv ~ $ rm -rf bfgminer
root@buildsrv ~ $ git clone https://github.com/luke-jr/bfgminer.git
Cloning into 'bfgminer'...
remote: Counting objects: 46656, done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 46656 (delta 2), reused 0 (delta 0), pack-reused 46650
Receiving objects: 100% (46656/46656), 36.48 MiB | 712 KiB/s, done.
Resolving deltas: 100% (29723/29723), done.
root@buildsrv ~ $ cd bfgminer
root@buildsrv ~/bfgminer $ ./autogen.sh
Getting submodules...
Submodule 'ccan' (git://git.ozlabs.org/~ccan/ccan) registered for path 'ccan-upstream'
Submodule 'knc-asic' (git://github.com/KnCMiner/knc-asic) registered for path 'knc-asic'
Submodule 'libbase58' (git://github.com/luke-jr/libbase58.git) registered for path 'libbase58'
Submodule 'libblkmaker' (git://gitorious.org/bitcoin/libblkmaker.git) registered for path 'libblkmaker'
Cloning into 'ccan-upstream'...
remote: Counting objects: 16804, done.
remote: Compressing objects: 100% (5225/5225), done.
remote: Total 16804 (delta 12183), reused 15554 (delta 11341)
Receiving objects: 100% (16804/16804), 12.74 MiB | 1.16 MiB/s, done.
Resolving deltas: 100% (12183/12183), done.
Submodule path 'ccan-upstream': checked out '8b0bdb090e2882aa431e89f4bc7aa4736e9e2838'
Cloning into 'knc-asic'...
remote: Counting objects: 898, done.
remote: Total 898 (delta 0), reused 0 (delta 0), pack-reused 898
[...]

PR #611, opened 2015-03-23
Fix to Sub Modules | Ensure Pool is sending block contents (acttx) before evaluating as fishy

daryltucker:
## Fix to Sub Modules

git:// protocol is typically used when explicit read/write/admin access has been granted for the user.  https:// protocol should be used for read-only use, such as for modules and dependencies.

Some users have set up configurations (ie: ssh) to connect to some hubs (ie: GitHub, BitBucket) with special keys.  This can cause complications when these users have not been granted read/write/admin access through git.
## Ensure Pool is sending block contents (acttx) before evaluating as fishy

Checking for server response before declaring mismatch.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,12 +1,12 @@
 [submodule "libblkmaker"]
-	path = libblkmaker
-	url = git://github.com/bitcoin/libblkmaker.git
+    path = libblkmaker
+    url = https://gitorious.org/bitcoin/libblkmaker.git
 [submodule "ccan"]
-	path = ccan-upstream
-	url = git://git.ozlabs.org/~ccan/ccan
+    path = ccan-upstream
+    url = git://git.ozlabs.org/~ccan/ccan
 [submodule "libbase58"]
-	path = libbase58
-	url = git://github.com/luke-jr/libbase58.git
+    path = libbase58
+    url = https://github.com/luke-jr/libbase58.git
 [submodule "knc-asic"]
-	path = knc-asic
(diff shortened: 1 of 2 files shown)


luke-jr:
https is not supported by all git clients (in particular, embedded systems often do not).

If acttx is wrong, then the pool is buggy and should be fixed - it isn't a BFGMiner problem.

daryltucker:
https://github.com/luke-jr/bfgminer/issues/608#issuecomment-85093838

daryltucker:
@luke-jr I agree with your statement:

```
If acttx is wrong, then the pool is buggy and should be fixed - it isn't a BFGMiner problem.
```

However, I differentiate between **wrong** and **not given**/**not supported**.  There isn't a mismatch; bfgminer is demanding something and if it doesn't get what it wants it says there is a mismatch.

If it were to say:

```
Pool %u is not sending us block contents
```

Would that be acceptable?  I think it's more clear, at least.  

BTW My miners didn't crash last night, after compiling with this.  Empirical, but it seems to help me :D

daryltucker:
@luke-jr 

Let's look at it another way.

If I patch [stratum-mining-proxy](https://github.com/slush0/stratum-mining-proxy), what do you want the response to be so that the user understands the difference between a mismatch and the pool's failure to provide the information at all?

Also, do you have any docs on what it is you actually want in that response?

luke-jr:
@daryltucker If pools do not support mining.get_transactions, they should return an error. That gets "Pool %u is not sending us block contents" instead.

PR #617, opened 2015-05-17
Copy kernels from sgminer

### LINKED ISSUE #429: Optimized scrypt kernels
Can BFGMiner include the optimized scrypt kernels that are packaged with sgminer?

vassilevsky:
Just copied the directory and added all non-conflicting files :)

This made me thinking: perhaps it would be better for the community if there was a single repository of just the kernels and then every miner author could add it as a submodule?

Closes #429


luke-jr:
You can't just copy kernels like this. The miner needs to actually support the algorithm and interface(s) too, or they won't work (see the malgo/ dir in BFGMiner). Right now, BFGMiner only has code for SHA256d, scrypt, and Keccak.

vassilevsky:
b7e

PR #624, opened 2015-07-14
Fix to enable building into directory out of source tree

DrYak:
This make it possible to "`../configure && make`" into a separate subdirectory.
(That's useful when building for multiple arches, each into its own subdirectory. Or simply to keep the source tree (relatively) clean)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -92,8 +92,8 @@ TEST_EXTENSIONS = .sh
 
 .PHONY: update-version
 update-version:
-	./gen-version.sh >version.h.new
-	cmp version.h version.h.new && rm version.h.new || mv version.h.new version.h
+	$(top_srcdir)/gen-version.sh >version.h.new
+	cmp $(top_srcdir)/version.h version.h.new && rm version.h.new || mv version.h.new $(top_srcdir)/version.h
 version.h: update-version
 bfgminer_SOURCES += version.c version.h
 BUILT_SOURCES = version.h
--- a/m4/bundled_lib.m4
+++ b/m4/bundled_lib.m4
@@ -50,7 +50,7 @@ AC_DEFUN([BFG_BUNDLED_LIB_VARS],[
 	BFG_CHECK_LD_ORIGIN
 	_AC_SRCDIRS(["$ac_dir"])
 	$1_CFLAGS='-I'"${ac_abs_top_srcdir}"'/$2'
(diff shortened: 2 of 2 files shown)


Makefile.am
```diff
@@ -92,8 +92,8 @@ TEST_EXTENSIONS = .sh
 
 .PHONY: update-version
 update-version:
-	./gen-version.sh >version.h.new
-	cmp version.h version.h.new && rm version.h.new || mv version.h.new version.h
+	$(top_srcdir)/gen-version.sh >version.h.new
+	cmp $(top_srcdir)/version.h version.h.new && rm version.h.new || mv version.h.new $(top_srcdir)/version.h
```
luke-jr, line 96: Should version.h be in the top_srcdir? Looks like it's more appropriate to put it in the build dir?

luke-jr:
@DrYak Hello?

DrYak:
Sorry for the slow answer.

Depends on what the data inside this file relates to (sorry I'm on my smartphone in a train, can't check now):
- Depends on the source code/on a particuliar release/github release version (I suspect that's the case), then it's similar to all the other release-specific data generated by autogen/automake/autoconf/libtool/etc. and should go into source_dir (as it was already the case. I've just added the necessary thing so that the file can be found when I'm in another directory)
- If it depends on a particular build (e.g.: for different architectures, or optimized vs debug configure, etc), like all the files generated by configure, then it should be moved, and the makefile changed so it can be found in the build_dir

luke-jr:
It's more of the former, but it is always generated at build-time.

DrYak:
Then I would leave it in source tree (i.e.: it doesn't change if you built for a different architecture, nor if you built against a different library, etc. It doesn't depend on the build configuration)

luke-jr:
Is it proper for make to modify the source tree, though?

DrYak:
Well, in theory, this could better go in autogen.sh

In practicie, given that lots of developer call autogen && configure from the make file (so that a configured build can update itself), and given that a makefile is the best place to catch missing files, I think it's okay.

luke-jr:
autogen.sh is not necessarily run on updates, but it's important to ensure version.h always has the correct git commit.

luke-jr:
Hmm, I wonder if gen-version.sh needs to be updated to change to top_srcdir as well. Thoughts?

luke-jr:
Yeah, fixing that in 6dab6fc0a2ce890542b19d87871ae3df7702a49c

PR #645, opened 2015-12-07
Cleaning conditional directives that break statements.

lzmths:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

```
https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/
```

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/driver-minergate.c
+++ b/driver-minergate.c
@@ -83,11 +83,13 @@ int minergate_open(const char * const devpath)
 	struct sockaddr_un sa = {
 		.sun_family = AF_UNIX,
 	};
+	bool test;
 #ifdef UNIX_PATH_MAX
-	if (devpath_len >= UNIX_PATH_MAX)
+	test devpath_len >= UNIX_PATH_MAX;
 #else
-	if (devpath_len >= sizeof(sa.sun_path))
+	test = devpath_len >= sizeof(sa.sun_path);
 #endif
+	if (test)
 		return -1;
 	const int fd = socket(PF_UNIX, SOCK_STREAM, 0);
 	strcpy(sa.sun_path, devpath);
--- a/util.c
+++ b/util.c
@@ -308,6 +308,7 @@ static int keep_sockalive(SOCKETTYPE fd)
 	const int tcp_keepidle = 45;
 	const int tcp_keepintvl = 30;
 	int ret = 0;
+	bool unlikely_test;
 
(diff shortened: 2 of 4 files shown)


driver-minergate.c
```diff
@@ -83,11 +83,13 @@ int minergate_open(const char * const devpath)
 	struct sockaddr_un sa = {
 		.sun_family = AF_UNIX,
 	};
+	bool test;
 #ifdef UNIX_PATH_MAX
-	if (devpath_len >= UNIX_PATH_MAX)
+	test devpath_len >= UNIX_PATH_MAX;
```
luke-jr, line 88: Pretty sure this is wrong?

driver-minergate.c
```diff
@@ -83,11 +83,13 @@ int minergate_open(const char * const devpath)
...
 #else
-	if (devpath_len >= sizeof(sa.sun_path))
+	test = devpath_len >= sizeof(sa.sun_path);
 #endif
+	if (test)
```
luke-jr, line 92: Wouldn't it be better to do:

``` C
#ifndef UNIX_PATH_MAX
    static const size_t UNIX_PATH_MAX = sizeof(sa.sun_path);
#endif
```

util.c
```diff
@@ -324,10 +325,11 @@ static int keep_sockalive(SOCKETTYPE fd)
 
 	if (!opt_delaynet)
 #ifndef __linux
-		if (unlikely(setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, (const void *)&tcp_one, sizeof(tcp_one))))
+		unlikely_test = unlikely(setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, (const void *)&tcp_one, sizeof(tcp_one)));
 #else /* __linux */
-		if (unlikely(setsockopt(fd, SOL_TCP, TCP_NODELAY, (const void *)&tcp_one, sizeof(tcp_one))))
+		unlikely_test = unlikely(setsockopt(fd, SOL_TCP, TCP_NODELAY, (const void *)&tcp_one, sizeof(tcp_one)));
 #endif /* __linux */
+		if (unlikely_test)
```
luke-jr, line 332: Doesn't the unlikely() need to be in the conditional?

luke-jr, line 332: Also, this is broken. Notice the conditional on opt_delaynet before the #ifndef

You'll need to add braces when breaking up the one-liner.

PR #662, opened 2016-05-18
Futurebit driver

tplink329242:


kwkr:
Is there a chance that this will be merged? I know it's old as f but it would be nice to have it in one place. Could I also somehow help in finishing it? I don't have any experience in this stuff, but maybe it could be a good place to make some first steps?

PR #666, opened 2016-06-20
Bugfix: gridseed: Implement a proper scanhash routine for GridSeeds

jeffrbeyer:
Previous behavior worked, but not intentionally


luke-jr:
?

PR #673, opened 2016-07-17
Updated Net Diff & Block stats...

GenTarkin:
To avoid confusion w/ multi-coin pools, Net Diff and Blocks stats now update to current information when "pool x issuing work for old block" message appears. Also, added "(Normal if multicoin pool)" message to the end of previously stated message.


luke-jr:
This breaks normal use. Multicoin pools simply need to use set_goal.

PR #717, opened 2017-09-17
modminer: Reset clocks to maximum when attempting to reduce below 178…

JohnnyPound:
… MHz


luke-jr:
I don't see the rationale behind this.

Issue #723, opened 2017-10-20
SIGSEGV in scrypt_1024_1_1_256_sp with musl


TheBiggerGuy:
When using Alpine Linux with musl libc on an ARM32v6 system
```
Thread 14 "miner_GSD0aa" received signal SIGSEGV, Segmentation fault.
[Switching to LWP 13764]
0xb6fb437c in memcpy () from /lib/ld-musl-armhf.so.1
(gdb) backtrace 
#0  0xb6fb437c in memcpy () from /lib/ld-musl-armhf.so.1
#1  0x00455210 in scrypt_1024_1_1_256_sp (input=0xb6b9ba4c, input@entry=0xb6b9ba6c,
    scratchpad=scratchpad@entry=0xb6b7b840 ":\204\332\070\061Spv,`\tH\266\063`k\216j\202\247\031\001\211\277\370t=\v\200\367\363\325\033%\355<-'q\203\262\376\252{\326\vI\343\071\304\023\065w\a\264\001v\006Zy\254\265\027\243\345\237\022L\236\203+i\"\246\240\f\262H\216&|\342{\nZ.\200
\226\307\064[\021\201\355D\216\022\347kn\371\300\247Ky\275HL\200\240f?~\031\312\037U\226\347\024\241\210\216_j\257%!\256X\261\333\302/A:w\352Hp\316\063\234\325#)\250\210m\263 \035\340_&\212HY\267\255\235\244ܟ\275i*z\241f$\276\346\344\001", <incomplete sequence \320>,
    ostate=0xb6b7b694, ostate@entry=0xb6b9ba4c) at malgo/scrypt.c:386
#2  0x00456df0 in scrypt_hash_data (out_hash=0x514a30, pdata=<optimized out>) at malgo/scrypt.c:502
#3  0x00416850 in work_hash (work=0x514970) at miner.c:10631
#4  _test_nonce2 (work=0x514970, nonce=<optimized out>, checktarget=<optimized out>) at miner.c:10656
#5  0x0042d688 in submit_noffset_nonce (thr=0x4984d0, work_in=0x54e950, nonce=3865654045, noffset=<optimized out>) at miner.c:10715
#6  0x0045d57c in gridseed_submit_nonce (thr=0x48df50, work=0xb6b9cb90, buf=0xb6b9cb98 <incomplete sequence \302>) at driver-gridseed.c:273
[...]

PR #724, opened 2017-10-20
Increase per thread stack size

TheBiggerGuy:
When using musl libc the default statck size is too small for the scrypt
code that uses alloca.

see: https://github.com/luke-jr/bfgminer/issues/723

### DIFF
--- a/miner.c
+++ b/miner.c
@@ -12609,9 +12609,17 @@ void start_cgpu(struct cgpu_info *cgpu)
 
 		thread_reportout(thr);
 
-		if (unlikely(thr_info_create(thr, NULL, miner_thread, thr)))
+		pthread_attr_t attr;
+		if (unlikely(pthread_attr_init(&attr)))
+			quit(1, "thread %d create failed (init)", thr->id);
+		if (unlikely(pthread_attr_setstacksize(&attr, (size_t) 2 << 20)))
+			quit(1, "thread %d create failed (set statck size)", thr->id);
+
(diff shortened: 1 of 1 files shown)


luke-jr:
Hmm, IIRC there is already stacksize stuff in BFGMiner. Do we want all these threads to have that stack size? Might it be a reduction in normal use cases, breaking other drivers?

TheBiggerGuy:
So first I should of made it clear with is a **Work In Progress** and should not be committed yet. The known issues with it are to do with page size alignment that I am still looking into.

To you questions:
 * "_already stacksize stuff in BFGMiner_": Didn't see any/can't find any, do you know where it is?
* "_Do we want all these threads to have that stack size_": This is only applied to the driver threads that do the scrypt hash and actually sets it to the GLibc default so is not worse than when not using musl libc.
* "_Might it be a reduction in normal use cases, breaking other drivers_": The `2 << 20` number is the GLibc default so I used that. To my page size alignment reference above the [libav bug](https://github.com/libuv/libuv/issues/1507) is better coded to read the current size and only bump it if required. I'm looking into that as a better patch.

For context I found this issues when moving to Alpine Linux to minimise a Docker container. This config is only compiled with scrypt, a single gridseed driver and on ARM.

luke-jr:
Looks like in the past we've just avoided using stack space too much.

Some #define magic may be needed for Windows: https://github.com/GerHobbelt/pthread-win32/blob/master/pthread_attr_setstacksize.c

PR #768, opened 2018-12-15
fix warning: checking address of array

isghe:
to avoid:
```
  CC       bfgminer-adl.o
adl.c:390:17: warning: address of array 'lpInfo[i].strAdapterName' will always evaluate to 'true' [-Wpointer-bool-conversion]
                if (lpInfo[i].strAdapterName)
                ~~  ~~~~~~~~~~^~~~~~~~~~~~~~
1 warning generated.
```

### DIFF
--- a/adl.c
+++ b/adl.c
@@ -385,9 +385,12 @@ void init_adl(int nDevs)
 		}
 
 		applog(LOG_INFO, "GPU %d %s hardware monitoring enabled", gpu, lpInfo[i].strAdapterName);
-		if (gpus[gpu].name)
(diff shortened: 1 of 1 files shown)


luke-jr:
```C
if (lpInfo[i].strAdapterName[0])
```
would probably be better

isghe:
> ```c
> if (lpInfo[i].strAdapterName[0])
> ```
> 
> would probably be better

Done; I had to `gpus[gpu].name = NULL;` because in that case it is a dangling pointer.

PR #840, opened 2022-06-04
Update .gitmodules (https:// instead of git://)

MarkSierra:
git:// does no longer work, therefore this file needs to be updated to https:// otherwise ./autogen.sh failes.


Frankbrids43 (commented):
That money belongs to me

PR #843, opened 2022-07-10
Intro.bfgminer

Atandanny:
BFGMiner driver configuration options and disable-other-drivers Build without drivers by default unless explicitly
	                        enabled
	--enable-broad-udevrules
	                        Include udev rules for ambiguous devices which may
	                        not be miners

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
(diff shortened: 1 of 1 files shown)


guvvy-tech:
Hey Dan:
may I call you by that?
Thanks much for the comeback
Oh, more confusion :O0
Might we converse via direct emaiL?
I think it might be beneficial to lay out what my HW environment is and describe what I know so- ie I know I am going to use CPU based- { I have enough cores on an ESX hypervisor - build 30+ 'rigs.'
pguvvy@gmail.com
Thank You
~~Paul

PR #850, opened 2023-06-18
submodules: Changing git to https protocol

Krellan:
GitHub has discontinued support for the git protocol:
https://github.blog/2021-09-01-improving-git-protocol-security-github/

Changing git:// to https:// accordingly.
Tested, and it now works, instead of timing out.


mvinijr:
Hello, I need BFG Miner because I'm studying Bitcoin node, could you give me a download link?

Eikyarkway (changes requested):
#850

PR #851, opened 2023-06-18
Adding the new SI prefixes for bigger units

Krellan:
Finally, they have added more prefixes beyond yotta:
https://www.bipm.org/en/cgpm-2022/resolution-3

Adding RH and QH for ronnahashes and quettahashes respectively.
Who knows, we might get there someday....


HTE9962:
hello

HTE9962:
why do downlaod the bfgminer?

